PR #1, opened 2019-04-04
fixed eslint errors

ivadimko:


src/arrayMethodReduce.test.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+const applyCustomReduce = require('./arrayMethodReduce');
+
+const source = [0, 10, 20, 30];
```
roman-mate: suggestion: sourceNumbers

PR #4, opened 2019-04-15
Ostap_Solution

ost-ptk:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let startValue = 0;
+    if (initialValue === undefined) {
+      initialValue = this[0];
```
ivadimko: I'd recommend avoiding overriding function arguments. Better to define a local variable along with `startValue` like `result = initialValue`

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
...
+      startValue = 1;
+    }
+    for (let i = startValue; i < this.length; i++) {
+      let element = this[i];
```
ivadimko: variable `element` is redundant. You can just pass `this[i]` in the callback

vadim-ilchenko (changes requested):

vadim-ilchenko (approved):

PR #7, opened 2019-05-01
sol

valeraonisko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,18 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let w = initialValue;
```
roman-mate: `w` is a bad name in this context. Perhaps `res` or `result`?

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,18 @@
...
+    this.forEach(function(el, i, thisArr) {
+      if (w === undefined) {
```
roman-mate: why do this inside of the loop?

PR #8, opened 2019-05-04
Solution

dvitalijb:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let resultOfMethod = initialValue;
```
roman-mate: just result. That's sufficient.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
...
+
+    if (initialValue === undefined) {
+      if (typeof this[0] === typeof '') {
```
roman-mate: why not 'string'?

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+
+    if (initialValue === undefined) {
+      if (typeof this[0] === typeof 'string') {
```
YuriiHoliuk: I think @roman-mate means `typeof this[0] === 'string'`

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
...
+        result = '';
+      } else if (typeof this[0] === typeof 1) {
```
YuriiHoliuk: same as above

PR #35, opened 2019-08-24
Vika Tsykun solution

ViktoriiaTsykun:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = arguments.length === 1 ? this[0] : initialValue;
```

PR #36, opened 2019-08-25
solution

Voronok16:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,15 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = this[0]) {
+    let redusedValue = initialValue;
```
irenhh: `reducedValue`

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,15 @@
...
+    if (typeof initialValue === 'string') {
```
irenhh: are you sure you need this?

Voronok16: Yes, without this condition, the string test ('takes first item as a `startValue` if not given, strings') fails. I had result:  "aabcd"

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,17 @@
...
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = this[0]) {
+    let reducedValue = initialValue;
+
+    if (typeof initialValue === 'string') {
+      reducedValue = '';
+    }
+    for (let i = 0; i < this.length; i++) {
```
irenhh: before this `for` you need to place blank line too

PR #37, opened 2019-08-25
Solution

VladyslavSak:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue === undefined ? undefined : initialValue;
```
irenhh: are you sure you need this condition?

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (accumulator !== undefined) {
+        accumulator = callback(accumulator, this[i], i, this);
+      } else {
+        accumulator = this[i];
+      }
+    }
+    return accumulator;
```
irenhh: new line before return

src/arrayMethodReduce.js
VladyslavSak: I can do it through the initialValue, but the linter doesn't miss me. So I decided so

irenhh: what linter says you, attach screenshot, please

VladyslavSak: image
But I disabled this error and added a new solution to GitHub.

irenhh: back to your prev solution, this is what was meant to edit
`let accumulator = initialValue === undefined ? undefined : initialValue;`
--->
`let accumulator = initialValue;`

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+
+    for (let i = 0; i < this.length; i++) {
+      if (accumulator !== undefined) {
```
danheim: а если мы захотим передать в качестве `initialValue` `undefined`?

PR #38, opened 2019-08-25
Solution

marchukmiroslava88:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this.length === 0) {
+      return initialValue;
+    }
+
+    let acc;
```

PR #40, opened 2019-08-26
Solution

kosbielozorov:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = 0) {
+    let i = 0;
+    let accumulator = initialValue;
+
+    if (arguments.length === 1) {
+      accumulator = this[0];
```
irenhh: instead of doing this, you can specify default value for `initialValue` and use ternary operator for `i`

irenhh (changes requested):

KosBelozyorov:
Fixed

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = 0) {
+    let i = 1;
+    let accumulator = initialValue;
+
+    arguments.length === 1
```
irenhh: 1. for one variable - one ternary operator
2. firstly you need to write the variable you define and then it's value or condition

irenhh (changes requested):

PR #41, opened 2019-08-26
Add solution

amelin0:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let resultReduce = initialValue;
+    if (arguments.length === 1) {
+      resultReduce = typeof this[0] === 'number' ? this[0] : '';
```
irenhh: array can consist of not only numbers

irenhh (changes requested):
set line breaks around different blocks of code

PR #42, opened 2019-08-26
fs_jul19 Alex Andrutskii solution

aleks-and:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = this[0];
```

irenhh (changes requested):

src/arrayMethodReduce.js
aleks-and: Fixed!

irenhh (approved):

PR #74, opened 2020-01-28
Implement method Reduce

cursorksu:


src/arrayMethodReduce.js
```diff
@@ -3,9 +3,25 @@
...
  */
+function isEqual(x1, x2) {
+  return x1 === x2 || (Number.isNaN(x1) && Number.isNaN(x2));
+}
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let counter = initialValue;
+    let beginOfIteration = 0;
```
irenhh:
```suggestion
    let startOfIteration = 0;
```
begin is used as a verb

PR #76, opened 2020-01-28
Sapuzhak: solution

Mikhail-88:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let value = initialValue;
+
+    for (let i = 0; i < this.length; i++) {
+      if (value === undefined) {
+        value = this[i];
```
irenhh: can you move this outside of loop and apply ternary operator?

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let value = initialValue === undefined ? this[0] : initialValue;
```
irenhh: it can be done directly in arguments
add a default value to arguments and remove this line

PR #78, opened 2020-01-30
solution

Valeriy-Sorochynskyi:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc;
+    let startIndex;
+
+    if (initialValue === undefined) {
```
irenhh: create default value for arguments

irenhh: and assign `initialValue` to `acc`

PR #130, opened 2020-06-01
Solution

campari229:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acumulator = initialValue;
+
+    if (acumulator === undefined) {
+      acumulator = this[0];
+    }
```
solaryasha: You could put this block in the `for` loop. I guess it would be better.

solaryasha (changes requested):

solaryasha (approved):

mriynyk (approved):

PR #154, opened 2020-08-10
Solution

aliceblunt8:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let returnValue = initialValue;
+    let startIndex = 0;
+
+    if (typeof initialValue === 'undefined' || initialValue == null) {
```
krajnova:
```suggestion
    if (initialValue === undefined || initialValue == null) {
```

krajnova: it would be shorter but it isn't necessary)

krajnova (approved):
looks nice)

src/arrayMethodReduce.js
aliceblunt8: thanks, fixed it)

irenhh (changes requested):
[1].reduce2()
this shouldn't work, check the original method

src/arrayMethodReduce.js
```diff
@@ -8,7 +8,11 @@ function applyCustomReduce() {
     let returnValue = initialValue;
     let startIndex = 0;
 
-    if (initialValue === undefined || initialValue == null) {
+    if (arguments.length < 1) {
+      return 'Uncaught TypeError: undefined is not a function at Array.reduce';
```
irenhh: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (arguments.length === 0 || typeof callback !== 'function') {
+      return;
```
irenhh: you return undefined if there is no callback,
according to the original method, you have an error, which is the result of calling the callback that doesn't exist;

just let this line `returnValue = callback(returnValue, this[i], i, this);` work and you will have the same errror

irenhh (approved):

PR #155, opened 2020-08-10
first solution

krajnova:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      start = 1;
+    }
+
+    for (let i = start; i < this.length; i++) {
+      const current = callback(accumulator, this[i], i, this);
+
+      accumulator = current;
```
aliceblunt8: мы можешь сохранять результат колбека в accumulator и тогда не нужна переменная current

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let start = 0;
+
+    if (arguments.length < 2) {
```
aliceblunt8: с arguments.length👍

aliceblunt8 (approved):

src/arrayMethodReduce.js
krajnova: Спасибо, пофиксила)

irenhh (changes requested):
[1].reduce2() this shouldn't work, check the original method

irenhh (changes requested):
you have another bug: initial value doesn't work as expected

irenhh (commented):
the first comment: bug again

irenhh (commented):
there is a problem with absence of the callback, forget about initial value for this case: [1].reduce2()

irenhh (approved):

PR #162, opened 2020-08-10
Solution

nznmv:


karinaKudatska (approved):
👍

irenhh (changes requested):
check how this [1].reduce2() works with the original method

nznmv:
> check how this [1].reduce2() works with the original method

честно говоря не до конца понял что именно не так ?

irenhh (approved):
now it works as expected 👍

PR #165, opened 2020-08-11
Solution

Andrii256:


Mikle25 (approved):

irenhh (changes requested):
[1].reduce2() check this with the origin method

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let startFrom;
+    let accumulator;
+
+    // in the case when callback havent got
+    if (arguments.length < 1) {
```
irenhh: this is a hack
to avoid it, you need to have startFrom as `0` when there is no callback provided

irenhh (approved):
💯

PR #167, opened 2020-08-11
solution

oivannikov:


nikita-parkhomenko (approved):

irenhh (changes requested):
check how this works with original method
[1].reduce2()

irenhh (approved):

PR #174, opened 2020-08-11
done

ChrisZinch:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+    let i = 0;
+
+    if (initialValue === undefined || initialValue === '') {
```
mpone: Думаю, що тут просто достатньо undefined.

mpone (changes requested):

mpone (approved):

irenhh (changes requested):
[1].reduce2()
this works not as expected, check the original method

src/arrayMethodReduce.js
```diff
@@ -5,6 +5,10 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
+    if (callback === undefined) {
+      return new TypeError('your callback undefined');
```
irenhh, line 9:
```suggestion
      return new TypeError('your callback is undefined');
```

irenhh (approved):

PR #178, opened 2020-08-11
Solution

mpone:


varseniuk (approved):
👍

irenhh (changes requested):
[1].reduce2()
check this for the original method and compare with yours

irenhh (approved):

PR #179, opened 2020-08-11
Solution

AnastasiiaBortnichuk:


leonbohdan (approved):

irenhh (changes requested):
[1].reduce2()
compare the result with the original method

irenhh (approved):

PR #180, opened 2020-08-11
solution

JuliaNastenko:


rmnkk (approved):
Nice👍

irenhh (changes requested):
what if callback is undefined as in [1].reduce2()?

irenhh (approved):

PR #181, opened 2020-08-11
solution

YuriiRysyn:


1nsider21 (approved):
Все вірно!

irenhh (changes requested):
check it for [1].reduce2() and compare with the original method

irenhh (approved):

PR #182, opened 2020-08-11
Solution

rmnkk:


JuliaNastenko (approved):

irenhh (changes requested):
what if callback is undefined as in [1].reduce2()?

irenhh (commented):
image

irenhh (approved):

PR #187, opened 2020-08-12
Solution

martakupka:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let previousValue = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
DmJSCourse, line 11: This can cause problems if additional arguments are added and therefore is (probably) not a best practice. It is better to check for initialValue explicitely, like (!initialValue) or (initialValue === undefined)

DmJSCourse (commented):
Everything works fine and very clear to read, only a minor stylistic problem present

mykola-yarchuk (approved):

PR #252, opened 2020-12-18
solution

troospieler:


VinnieJ-2k20 (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result;
+
+    if (initialValue || initialValue === 0) {
+      result = initialValue;
+
+      for (let i = 0; i < this.length; i++) {
+        result = callback(result, this[i], i, this);
+      }
```
varseniuk: DRY. You use two absolutely identical 'for' loops, which differ only with the start index (0 or 1).
Let's try to make your code smaller and more simple.
So all you need - it's to run the check if 'initialValue' is defined or it is undefined. According to this, you can find out, which value to assign to your 'result' variable (what you have already done in your code) and which start index you need  (it depends on the case if 'initialValue' is defined or undefined). You may create another variable for this e.g. 'startIndex' and put it in the loop conditions `for (let i = startIndex; i < this.length; i++)`
And when you have your 'startIndex' and 'result', you can run only one 'for' loop instead of two

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
...
+    }
+
+    if (!initialValue && initialValue !== 0) {
```
```suggestion
    if (!initialValue) {
```

unnecessary another check. If there is NO initialValue, why do you also check if it's !==0?

varseniuk (changes requested):

mgrinko (approved):

PR #261, opened 2020-12-18
solution

dtsehelnyk:


Pasha5555 (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
-    // write code here
+    let acc;
+    let loopStart = 0;
+
+    if (initialValue !== undefined) {
+      acc = initialValue;
+    } else {
+      acc = this[0];
+      loopStart++;
+    }
```
```suggestion
    let acc = initialValue;
    let loopStart = 0;

    if (initialValue === undefined) {
      acc = this[0];
      loopStart = 1;
    }
```

The logic is correct, but let's make your code clearer and get rid of unnecessary 'else'.
You need to change your 'acc' and 'loopStart' variables only if 'intialValue' is undefined. Only one check.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc;
+    let loopStart = 0;
+
+    if (initialValue !== undefined) {
+      acc = initialValue;
+    } else {
+      acc = this[0];
+      loopStart++;
```
varseniuk: there's no need to use incrementation here. You need only either 0 or 1. Nothing more

varseniuk (changes requested):
Nice, but need some corrections

varseniuk (approved):

PR #282, opened 2020-12-21
Solution

Dima-Semenov:


kerpichonka (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex++;
```
varseniuk, line 13:
```suggestion
      startIndex = 1;
```

No need to use incrementation here. There are only two values you need - either 1 or 0 (depending on arguments)

varseniuk (approved):
Nice.
Pay attention to my comment

src/arrayMethodReduce.js
Dima-Semenov: Зрозумів

PR #364, opened 2021-04-26
Solution

Burunduk003:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,22 @@
...
     // write code here
+    let startIndex = 0;
+    let accumulator = initialValue;
+
+    if (initialValue === undefined) {
+      accumulator = this[0];
+      startIndex = 1;
+    }
+
+    let currentValue;
```
igor-stasiv, line 17: You don't need this variable, you can just use this[i]. By the logic of the code, it is clear that this[i] is the current value

igor-stasiv (approved):
Good job!

src/arrayMethodReduce.js
mykola-yarchuk, line 17: agree with @igor-stasiv

mykola-yarchuk (approved):
great

PR #365, opened 2021-04-26
Solution

igor-stasiv:


Burunduk003 (approved):
Good job! Naming could be more specific but for such a small task, it will do.

varseniuk (approved):
Great job 🔥

PR #367, opened 2021-04-28
made custom reduce

Ihor-Karpyn:


nazarmatsevych (approved):
Nice work! 👍

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let start = 0;
+    let feedBack = initialValue;
+
+    if (arguments.length === 1) {
```
mykola-yarchuk, line 11:
```suggestion
    if (initialValue === undefined) {
```
your version works well but is better when your condition strict

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let start = 0;
+    let feedBack = initialValue;
```
mykola-yarchuk, line 9: I'm not sure about feedback, usually, it called an accumulator.

mykola-yarchuk (approved):
great

PR #378, opened 2021-04-29
Solution-001

senkiv-oleh:


dmytro-abram (approved):
Хороша робота

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
...
+    let accumulator = initialValue;
+    let index = 0;
+
+    if (initialValue === undefined) {
+      accumulator = this[0];
+      index = 1;
+    }
+
+    while (index < this.length) {
+      accumulator = callback(accumulator, this[index], index, this);
```
varseniuk, line 17: This can be a bit shorter if use `for` loop, but your solution is ok as well

varseniuk (approved):
Good job 🔥

PR #384, opened 2021-04-29
Reduce method

leonid-vegera:


IvanKolodiazhnyi (commented):
Good job)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let summary = initialValue;
```
varseniuk, line 8:
```suggestion
    let accumulator = initialValue;
```
In fact, this is an accumulator - it accumulates all previous values. It's more meaningful name

varseniuk (approved):
Great! 👍

PR #385, opened 2021-04-29
Solution

IvanVasylenko:


ivan-shpynda (approved):
good!

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
```
```suggestion
    let accumulator = initialValue;
```
This variable 'accumulates' all results, so it would be better to call it in such way

varseniuk (approved):
Good! 👍

PR #386, opened 2021-04-29
Solution

SergeyZub:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,31 @@
...
+        result += this[i];
+      }
+
+      if (callback.length === 3) {
+        result += i;
+      }
+
+      if (callback.length === 4) {
+        result = this.length;
+      }
```
IvanKolodiazhnyi: Я не докінця розумію для чого?
Ти міг просто написати так result = callback(result, this[i], i, this);

SergeyZub: согласен не туда понесло меня

varseniuk: Ivan is right - these checks are redundant here. Let's follow this advice and make this callback apply 4 arguments (as the original `reduce` method does)

varseniuk (changes requested):
Let's follow the advice of your colleague and make the code more simple and meaningful

SergeyZub:
Уже изменил

varseniuk (approved):
Nice 👍

PR #387, opened 2021-04-29
solution

TimurHrynenko:


nazarmatsevych (approved):

aliceblunt8 (changes requested):
- let's assign default values to startIndex and only if their is no initialValue reassign them
- no need for if else here
- also there should be only ONE for-loop

aliceblunt8 (approved):

PR #391, opened 2021-04-29
solution

Allrightman:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = initialValue;
```
IvanKolodiazhnyi: I don't understand this name of variable(
Code is nice)
Good job)

Allrightman: Thanks!
"acc" - it's short from "accumulator" - reduce(accumulator, currentValue),
that's the same what Misha writes "prev"

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+
+    let i = 0;
```
aliceblunt8:
```suggestion
    let startIndex = 0;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+
+    if (arguments.length < 2) {
+      acc = this[i];
+      i++;
```
aliceblunt8:
```suggestion
      startIndex = 1;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+    }
+
+    while (i < this.length) {
+      acc = callback(acc, this[i], i, this);
+      i++;
+    }
```
aliceblunt8: use for loop

src/arrayMethodReduce.js
aliceblunt8: let's rename to result or returnValue

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #399, opened 2021-05-03
solution

RuslanaPodshyvalova:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,34 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this.length === 0 && initialValue === undefined) {
+      throw new TypeError('array is empty');
+    }
```
aionova:
```suggestion
    if (this.length === 0 && initialValue === undefined) {
      throw new TypeError('Reduce of empty array with no initial value at Array.reduce');
    }
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,34 @@
...
+
+    if (this.length === 1 && initialValue === undefined) {
+      return this.filter(item => item !== undefined);
+    }
+
+    if (this.length === 0 && initialValue !== undefined) {
+      return initialValue;
+    }
+
```
aionova: no need for these checks

aionova (changes requested):
Looks good, but let's refactor the code to make it shorter

aionova (approved):
Nice!

PR #400, opened 2021-05-04
reduce

YuriiLiso:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let lastResult = initialValue;
+    let i = 0;
+
+    if (arguments.length < 2) {
```
vova-myronenko: I don't know which way is better whether to check initialValue for equality to undefined or to check arguments length. Or maybe it's not an issue at all. I hope mentors can clarify it.

vova-myronenko: Anyway, good job!

vova-myronenko (approved):

vova-myronenko:
Anyway, good job!

src/arrayMethodReduce.js
VinnieJ-2k20: This method also works, but it is definitely better to check if `initialValue === undefined`. This shows the exact logic of your code.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let lastResult = initialValue;
+    let i = 0;
```
VinnieJ-2k20: Use separate variables for logic and loops.

```suggestion
    let startIndex = 0;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+
+    if (arguments.length < 2) {
+      lastResult = this[0];
+      i = 1;
+    }
+
+    for (i; i < this.length; i++) {
```
VinnieJ-2k20: See top comment

```suggestion
    for (let i = startIndex; i < this.length; i++) {
```

VinnieJ-2k20 (changes requested):
Almost finished. Some changes required.

VinnieJ-2k20 (approved):
Good job 👍

PR #401, opened 2021-05-04
Solution

MinchukO:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
VinnieJ-2k20: In this case more reliable check would be 
```suggestion
    if (initialValue === undefined) {
```

VinnieJ-2k20 (changes requested):
Solution is correct, just fix some logic details.

VinnieJ-2k20 (approved):

PR #409, opened 2021-05-08
solution

dmitry-puhliakov:


src/arrayMethodReduce.js
```diff
@@ -3,9 +3,17 @@
 /**
  * Implement method Reduce
  */
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = (initialValue === undefined) ? this[0] : initialValue;
+    const start = (initialValue === undefined) ? 1 : 0;
```
varseniuk, line 10: DRY. Here you have two similar checks `initialValue === undefined`.
You may use it only once like this:
```
  let accumulator = this[0];
  const start = 1;
} else {
  const start = 0;
}
```

varseniuk (approved):
Good job!
But you can make it even better. Read my comment

PR #472, opened 2021-08-11
Solution

sasha-morozov:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
```
DmytroMykoliv: I had comments on abbreviated variables, I advise you to write instead prev with previous )

DmytroMykoliv (approved):
Looks good )

src/arrayMethodReduce.js
sasha-morozov: Ok, will fix this one!)

herkoss (approved):
Well done!✅👏

PR #473, opened 2021-08-11
JS Advanced - Reduce method

george-w-boo:


sasha-morozov (approved):
Good solution!)

herkoss (approved):
Well done!👍👏

PR #475, opened 2021-08-12
it's work

evgen1sider:


DmytroMykoliv (approved):
good solution ))

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result;
```
VinnieJ-2k20: You can assign `initialValue` here from the start and avoid `else` later
```suggestion
    let result = initialValue;
```

VinnieJ-2k20 (approved):
Good job 👍

PR #480, opened 2021-08-12
solution

ENV-S:


AndreySerhovetc (approved):

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,20 @@
...
+
+    if (initialValue === undefined) {
+      result = this[0];
+      index = 1;
+    }
+
+    while (index < this.length) {
+      result = callback(result, this[index], index, this);
+      index++;
+    }
```
VinnieJ-2k20: `for(...)` loop would be appropriate here.
Thumb rule is: 
1. If you know the exact number of iterations (like in this case) - use `for` with index
2. If you don't know the exact number of iterations, or it depends on some other conditions - use `while` or `do...while`.

VinnieJ-2k20 (approved):
Good job 👍 
Read the comment on some advice on loops.

PR #481, opened 2021-08-12
solution reduce done

AndreySerhovetc:


maximbalyk (approved):
good job

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
VinnieJ-2k20: This check is not very clear. use more obvious one
```suggestion
    if (initialValue === undefined) {
```

VinnieJ-2k20 (changes requested):
Solution works, but need to change one small detail.

src/arrayMethodReduce.js
AndreySerhovetc: fixed

VinnieJ-2k20 (approved):
Good job 👍

PR #485, opened 2021-08-12
Array method reduce

kristof22122:


Andrii-Pinchuk (approved):
Good job 👍

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let previous = initialValue;
+    let start = 0;
+
+    if (arguments.length < 2) {
```
VinnieJ-2k20: this is not an obvious check. Let's use more appropriate condition here:
```suggestion
    if (initialValue === undefined) {
```

VinnieJ-2k20 (changes requested):
Solution is correct, but you need to fix one important detail.

VinnieJ-2k20 (approved):
Good job 👍

PR #491, opened 2021-08-13
solution

OleksandrDzindziura:


sasha-morozov (approved):
Good job!

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
VinnieJ-2k20: Let's use more obvious check here.
```suggestion
    if (initialValue === undefined) {
```

VinnieJ-2k20 (changes requested):
Almost done, just fix one important thing.

VinnieJ-2k20 (approved):
Good job 👍

PR #497, opened 2021-08-15
solution

RomSmile:


AndreySerhovetc (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,27 @@
...
+    let startIndex;
+
+    if (initialValue === undefined) {
+      if (typeof this[0] === 'string') {
+        startIndex = 1;
+        newInitialValue = this[0];
+      } else {
+        startIndex = 0;
+        newInitialValue = 0;
+      }
```
VinnieJ-2k20, line 18: We really don't need type check here. It only depends on existence of `initialValue`.

VinnieJ-2k20 (changes requested):
Almost done, one small thing to fix.

PR #548, opened 2021-10-15
add check for undefined as initial value

solaryasha:


VinnieJ-2k20 (approved):
You are a very cruel person 😈

PR #576, opened 2021-11-15
Reduce method

Vitaliy-0:


SerhiiKirik (approved):
Good job 🔥

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      startIndex++;
+      prev = this[0];
+    }
```
ihor-jpeg, line 14: can you think of a way of getting rid of this block?

ihor-jpeg (approved):
Good work!

src/arrayMethodReduce.js
Vitaliy-0: no, help

PR #578, opened 2021-11-15
Solution

alinaShchavinska:


alex-kostomarov (approved):
Good job! 🔥

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, startValue) {
+    let previousValue = startValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      previousValue = this[0];
```
ihor-jpeg, line 12: you can set this as a default value of the start value parameter

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
...
+      startIndex = 1;
```
ihor-jpeg, line 13: you can use ternary to set the value of the start index and you wont need this if block

ihor-jpeg (approved):
Good work.

PR #581, opened 2021-11-16
init

ghost:


msaikun (approved):
LGTM!

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,22 @@
...
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function (callback, initialValue) {
+    let accum = initialValue;
+    let beginIndex = 0;
+
+    if (arguments.length === 1) {
+      accum = this[0];
+      beginIndex = 1;
+    }
```
ihor-jpeg, line 14: it's a bit strange that first you assign a possibly undefined value to a variable and only after that you check the arguments.length

ihor-jpeg (approved):
Good work. Have a look at the comments

PR #599, opened 2021-11-19
solution

lanebx:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let tempResultCallback = initialValue;
+
+    if (arguments.length === 1) {
+      tempResultCallback = this[0];
+
+      for (let i = 1; i < this.length; i++) {
```
lera-permiakova: You can create a variable for the start index and do without second loop

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let tempResultCallback = initialValue;
+
+    if (arguments.length === 1) {
+      tempResultCallback = this[0];
```
lera-permiakova: I think, it's better to do it in one line using ternary operator

lera-permiakova (approved):
GJ!

AlinaMatuschak (changes requested):
fix @lera-permiakova comments

AlinaMatuschak (approved):

PR #601, opened 2021-11-19
Solution of array-method-reduce

Dima-Ziakhor:


alinaShchavinska (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result;
```
AlinaMatuschak: you can take a better name

AlinaMatuschak: you can use logical operand
```suggestion
    let result = initialValue || this[0];
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+    let startIndex = 0;
```
AlinaMatuschak: there is better use ternary operator
if an initial value exists return 0. If not - return 1

AlinaMatuschak (changes requested):
Almost perfect 👍

src/arrayMethodReduce.js
Dima-Ziakhor: I thought to do the same, but some tests fail because of this(

AlinaMatuschak: What test?

AlinaMatuschak (approved):
🔥

PR #605, opened 2021-11-20
Develop

WorkAccCom:


lera-permiakova (approved):
GJ!

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prevCallResult;
```
AlinaMatuschak, line 8: you can use logical operand
```suggestion
    let prevCallResult = initialValue || this[0];
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
+    let i = 0;
```
AlinaMatuschak, line 9: there is better use ternary operator
if an initial value exists return 0. If not - return 1

AlinaMatuschak (approved):
Nice work 👌

src/arrayMethodReduce.js
WorkAccCom: @AlinaMatuschak but what if undefined explicitly passed to an initial value?

PR #667, opened 2022-02-01
Reduce method

polosanya:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,15 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const onlyOneArg = arguments.length === 1;
+    let accumulator = onlyOneArg ? this[0] : initialValue;
+    const currentValue = onlyOneArg ? 1 : 0;
```
Anna-Tkachenko, line 10: It's not value, it's start index

Anna-Tkachenko (approved):

PR #721, opened 2022-04-06
Create solution

ivan-kadykalo:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    } else if (typeof initialValue === 'undefined') {
+      prev = `undefined`;
+    }
```
YegorKochetkov: extra check, it will work without it
 if 'undefined' it already is in prev (at line 8)

YegorKochetkov (changes requested):

AlinaMatuschak (approved):

PR #722, opened 2022-04-06
Initial commit - solution

paralele-piped:


src/arrayMethodReduce.js
```diff
@@ -3,10 +3,40 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const startValue = arguments.length <= 1
+      ? 1
+      : 0;
+
+    let previousValue = arguments.length <= 1
+      ? this[0]
+      : initialValue;
```
Bohdan-Maksymiuk: you did two similar checks - unite them

src/arrayMethodReduce.js
```diff
@@ -3,10 +3,40 @@
...
+
+    for (let i = startValue; i < this.length; i++) {
```
Bohdan-Maksymiuk:
```suggestion
    for (let i = startIndex; i < this.length; i++) {
```
it's index, not value

Bohdan-Maksymiuk (changes requested):

src/arrayMethodReduce.js
paralele-piped: Changed code, check it out

src/arrayMethodReduce.js
```diff
@@ -3,9 +3,25 @@
...
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let startIndex;
+    let previousValue;
+
+    if (arguments.length <= 1) {
+      startIndex = 1;
+      previousValue = this[0];
```
ArtemTopolskyi, line 14: you can use default parameter:
function(callback, initialValue = this[0])
but it's also ok

ArtemTopolskyi (approved):

PR #747, opened 2022-04-16
reduce

Nikitiliok:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+    let i = 0;
+
+    if (arguments.length < 2) {
+      i = 1;
+      result = this[0];
+    }
+
```
Yurii837, line 15: You can use additional variable, but i think it's nice solution.

Yurii837 (approved):
Interesting solution

MaksimusCRS (approved):

PR #752, opened 2022-04-29
Solution

59MrRobot:


LukaszGryziewicz (approved):
🥇

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
```
ihor-jpeg, line 14: can we do it without if and with const?

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #759, opened 2022-05-03
solution

AnastasiyaSopotnytska:


vlasiuk-anatolii (approved):
Good Jod!

Anna-Tkachenko (approved):
Good job! ⭐️
Just use previousValue as variable name instead of prev)

PR #767, opened 2022-05-04
task completed

andreisucman:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
amitpatiljc: Good work ! 
Just a suggestion I got from Paul 
that , Try to do such assignments using ternary operators.

   const prev = (arguments.length < 2)  ?  this[0]  :  initialValue;
   const startIndex = (arguments.length < 2)  ?  1  :  0;

amitpatiljc (approved):

amitpatiljc (approved):
Good Job ! Just one minor suggestion.

src/arrayMethodReduce.js
ansu16281: Okay, will do.

VinnieJ-2k20: Not done :)

VinnieJ-2k20 (changes requested):

src/arrayMethodReduce.js
ansu16281: I committed it yesterday as well as the other task but it doesn't show up.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = arguments.length < 2 ? this[0] : initialValue;
+    const startIndex = arguments.length < 2 ? 1 : 0;
```
ihor-jpeg:
```suggestion
    const startIndex = arguments.length < 2
      ? 1
      : 0;
```

ihor-jpeg (approved):

PR #768, opened 2022-05-04
solutionI

EnjSadman:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,42 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let accum;
+
+    if (arguments.length === 1) {
+      if (typeof (this[0]) === 'string') {
```
SavMary: There is no need to check the type of array elements. If initialValue is passed, accum = initialValue, otherwise accum = start value of the array. 
Also, here you can find the index for pass to a callback at once and you will have all the necessary parameters.

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,42 @@
...
+        accum = '';
+      } else if (typeof (this[0]) === 'number') {
+        accum = this[0];
+      }
+    } else {
+      accum = initialValue;
+    }
+
+    switch (callback.length) {
```
SavMary: If you got all parameters, you can pass them immediately without construction switch-case

SavMary (changes requested):

denys-cheporniuk (approved):

PR #772, opened 2022-05-06
solution

huberciak:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result;
+    let start;
+
+    if (arguments.length === 2) {
+      result = initialValue;
+      start = 0;
+    } else {
```
efdalyalcin: I think you can give values to the variables where you introduce. Then you will not need the else statement, which makes code harder to read.

efdalyalcin (changes requested):
Great job except the little comment.

efdalyalcin (approved):
well done, clean code!!

meforoma (approved):

PR #773, opened 2022-05-06
solution

SavMary:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const arr = Object(this);
```
vlad-demchuk: again this 'arr')

vlad-demchuk (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
+    let value;
+    let index;
+
+    if (arguments.length > 1) {
+      value = initialValue;
+      index = 0;
+    } else {
+      value = this[0];
+      index = 1;
+    }
```
denys-cheporniuk, line 17: better to check `arguments.length < 2` and remove else block

denys-cheporniuk (approved):

PR #774, opened 2022-05-06
solution

yelnikov-andrii:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let res = initialValue;
```
SavMary: You should write full names of variables for better understanding

SavMary (approved):

src/arrayMethodReduce.js
denys-cheporniuk: agree, create better name for a variable

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+      for (let i = 0; i < this.length; i++) {
+        res = callback(res, this[i], i, this);
+      }
+    } else {
+      res = this[0];
+
+      for (let i = 1; i < this.length; i++) {
+        res = callback(res, this[i], i, this);
+      }
+    }
```
denys-cheporniuk: make `if` before loops and use only one loop, you don't need two loops here

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #809, opened 2022-06-10
my solution

Pavlo000:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
+
+    for (startIndex; startIndex < this.length; startIndex++) {
+      prev = callback(prev, this[startIndex], startIndex, this);
+    }
```
stanislav-topikha: it's better to use separate variable for Index
`for (let index = startIndex; index < this.length;  index++) {`

stanislav-topikha (changes requested):

src/arrayMethodReduce.js
stanislav-topikha: Else ok 👍

Pavlo000: corrected

stanislav-topikha (approved):

mkyloa (approved):

PR #810, opened 2022-06-11
Solution

senile-pitcher:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const thisArray = [...this];
+
+    if (thisArray.length === 0 && initialValue === undefined) {
+      throw new TypeError('reduce of empty array with no initial value');
+    }
```
AndriiNychaiuk, line 14: Great solution! It's easy and simple to understand!) But there is one moment. There won't be an error in the reduce method if the array is empty and the initial value is undefined. So you don't need this part in your solution I guess.

AndriiNychaiuk (changes requested):

AndriiNychaiuk (approved):
👍

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const thisArray = [...this];
```
G-YaSha: Why do need this copy? You don't mutate `this`, so you don't need copy here

src/arrayMethodReduce.js
G-YaSha: Actually, It will be the error in such case, so good job👍

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,29 @@
...
+
+    if (callback === undefined) {
+      throw new TypeError('undefined is not a function');
+    }
```
G-YaSha, line 10: 👍

G-YaSha (changes requested):
Almost done🔥 Great solution, let's fix some small thing

G-YaSha (approved):

PR #814, opened 2022-06-12
all tests passed

frontend112:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,37 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = '') {
```
ihor-jpeg: but the initial value, if not passed, should not be an empty string. it should be the first element of the array

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,37 @@
...
+    let sum = initialValue;
+    // if given array contains only strings
+    let startValue = '';
+
+    if (typeof this[0] === 'number') {
+      startValue = 0;
+    }
+
+    // when passed two arguments and equal to undefined
+    if (arguments.length === 2 && arguments[1] === undefined) {
```
ihor-jpeg: the second argument will never be undefined unless you pass it explicitly, which you never wanna do. you have set the default value for `initialValue`, that's why it will never be undefined

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,37 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = '') {
+    let sum = initialValue;
+    // if given array contains only strings
```
ihor-jpeg: what if it doesn't?

ihor-jpeg (changes requested):
your code is a bit imperative, it checks strictly for numbers or strings, but you should not care about what an element is. it can be an array of objects.

- if initialValue is not passed, it should be the first element of the array
- callback for every item
- the end

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let sum = initialValue;
```
meforoma, line 8: more of a semantic thing - `.reduce` as a method returns an `accumulator` that might be any data type - not necessarily sum (yet, yes - sum is the most often usage)
so just from the naming perspective it's `accumulator` or `result`

meforoma (approved):
concise and effective 👍

PR #818, opened 2022-06-13
Solution

ostapfriend:


vsupruniuk (approved):
Good solution👍

G-YaSha (approved):
You should commit only files, you've changed not all.

PR #828, opened 2022-06-14
first version

DmitriiPirohov:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
...
+
+    if (arguments.length < 2) {
+      sum = this[0];
+      count++;
+    }
+
+    while (count < this.length) {
+      sum = callback(sum, this[count], count, this);
+      count++;
+    }
```
G-YaSha: Use `while` loop when you don't now number of iterations, in this better to use just `for` loop

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let sum = initialValue;
```
G-YaSha: It does not always sum:
```suggestion
    let acumulator = initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
...
+    let count = 0;
```
G-YaSha: Count not the best name here, it's rather `startIndex`, not count

G-YaSha (changes requested):

jstmpelowycz (approved):

PR #831, opened 2022-06-14
solution

AKomashchuk:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
-    // write code here
+    let previousValue = initialValue;
+    let startWith = 0;
+
+    if (arguments.length < 2) {
+      previousValue = this[0];
+      startWith = 1;
+    }
+
+    for (let i = startWith; i < this.length; i++) {
```
Victance, line 16: This is about English not code

it is better to use 'startFrom', 'startAt' or 'startIndex'  instead of 'startWith'
check the difference between them https://english.stackexchange.com/questions/112972/starting-with-vs-starting-from

Victance (approved):
Nicely done🔥

mkyloa (approved):

PR #833, opened 2022-06-15
Solution

Victance:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    if (arguments.length < 2) {
+      i = 1;
+      prev = this[0];
+    }
+
+    for (i; i < this.length; i++) {
+      prev = callback(prev, this[i], i, this);
+    }
+
+    return prev;
```
G-YaSha: Let's improve your solution a bit
```suggestion
    let prev = initialValue;
    let startIndex = 0;

    if (arguments.length < 2) {
      startIndex = 1;
      prev = this[0];
    }

    for (let i = startIndex; i < this.length; i++) {
      prev = callback(prev, this[i], i, this);
    }

    return prev;
```

G-YaSha (changes requested):

G-YaSha (approved):

PR #879, opened 2022-07-12
Oleksandr-Pustovit

Oleksandr-Pustovit:
Прекрасний прилад того як нормально працюючу функцію слід підлаштовувати під юЗВІРів

### DIFF
--- a/src/arrayMethodReduce.js
+++ b/src/arrayMethodReduce.js
@@ -5,28 +5,28 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, ...initialValue) {
-    let initialValue1 = initialValue;
+    let initial = initialValue;
 
-    if (initialValue1.length === 0) {
-      initialValue1 = undefined;
+    if (initial.length === 0) {
+      initial = undefined;
     } else {
-      if (initialValue1[0] === undefined) {
-        initialValue1 = 'undefined';
+      if (initial[0] === undefined) {
+        initial = 'undefined';
       } else {
-        initialValue1 = initialValue1[0];
+        initial = initial[0];
       }
     }
 
     if (this.length < 1) {
-      return initialValue1;
+      return initial;
     }
 
     let first;
     let second;
     let index = 0;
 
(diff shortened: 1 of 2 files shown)


src/arrayMethodReduce.js
```diff
@@ -5,28 +5,28 @@
...
+        initial = initial[0];
       }
     }
 
     if (this.length < 1) {
-      return initialValue1;
+      return initial;
     }
 
     let first;
```
blcts: I think its better to rename

blcts (changes requested):

src/arrayMethodReduce.js
Oleksandr-Pustovit: I think it is better to be able to reuse parameters of the function. Especially if this task is a clear abuse without knowledge about agruments and length. (which are mentioned later on the course).
To sum up I am in need to create a new vavriable with simple name `initialValue` without using `initialValue`. I am sure `initial` is much better than `initialV | iValue | inVal` and the one which describes it the most `valueWhichCanBeUndefined`

blcts (approved):
good :)

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,43 @@
...
+
+    if (initial.length === 0) {
+      initial = undefined;
+    } else {
+      if (initial[0] === undefined) {
+        initial = 'undefined';
+      } else {
+        initial = initial[0];
+      }
+    }
```
ivan-kadykalo: Avoid using nested if statements when it is possible. I'm sure you can do your code without nesting🙂

ivan-kadykalo (changes requested):

src/arrayMethodReduce.js
Oleksandr-Pustovit: I am sure users can avoid using undefined as values or if they want `'undefined'` they will use `'undefined'` not `undefined`.
But I will change the aproach here.

ivan-kadykalo (approved):

PR #889, opened 2022-07-13
solution

AleksandrYuschenko:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,22 @@
...
+
+    let result = initialValue;
+    let index = 0;
+
+    if (initialValue === 'undefined') {
+      result = 'undefined';
+    } else if (arguments.length === 1) {
+      result = this[0];
+      index = 1;
+    }
```
oleksandr-kotliarov: or just

```suggestion
    if (arguments.length < 2) {
      prev = this[0];
      startIndex = 1;
    }
```

oleksandr-kotliarov (approved):
good

src/arrayMethodReduce.js
Ihor-Karpyn: agree, let's fix it

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #895, opened 2022-07-14
Solution

illnino380:


Vova-Balyshev (approved):
База!

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
...
-    // write code here
+    let accumulator = initialValue;
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === 0
+          && initialValue === undefined
+          && arguments.length === 1) {
+        accumulator = this[0];
+        continue;
+      }
```
ivan-kadykalo: This check will work on each loop but you really need id only once at the start. Maybe will be better to move it outside of your for loop.

ivan-kadykalo (changes requested):

Ihor-Karpyn (approved):

PR #898, opened 2022-07-15
Implementation reduce

VolodymyrShymanskyi:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let sum = initialValue;
+    let i = 0;
+
+    if (arguments.length === 1) {
+      sum = this[0];
+      i = 1;
+    }
```
amitpatiljc: Try to do such assignments using ternary operators.
and try to give meaningful variable names 
for example instead of "i" you can give "startIndex"

amitpatiljc:
```suggestion
    const sum = (arguments.length === 1) ? this[0] : initialValue;
    const i = (arguments.length === 1)  ? 1 : 0;
```

amitpatiljc (changes requested):
Good job 👍
Added one comment.

src/arrayMethodReduce.js
VolodymyrShymanskyi: The changes were maden

amitpatiljc (approved):
Well Done

PR #915, opened 2022-07-31
Solution

denys-sadovyi:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+    let y = 0;
```
valerii-derkach, line 9: better to name variables meanful 
`initialValue`, for example

valerii-derkach (approved):

PR #930, opened 2022-08-11
First commit

dm-sheremetiev:


ArturErmolenko (approved):

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,22 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = 'undefined') {
```
Bohdan-Maksymiuk, line 7: looks weird to set initialValues as  `undefined` string by default
and taking into consideration that you have check on arguments length then

Bohdan-Maksymiuk (changes requested):

PR #931, opened 2022-08-11
solution

AndreyJigijalkin:


OlehMy (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let value = arguments.length > 1
+      ? initialValue : this[0];
```
```suggestion
    let value = arguments.length > 1
      ? initialValue
      : this[0];
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
...
+
+    const iterations = arguments.length > 1
+      ? 0 : 1;
```
AlinaMatuschak: the same

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
...
+
+    for (let i = iterations; i < this.length; i++) {
+      value = callback(value, this[i], i, this);
+    }
```

AlinaMatuschak (changes requested):
Nice 👍 , just use array method

src/arrayMethodReduce.js
AndreyJigijalkin: I am not sure if I can use array methods in this task, because on MA platform in all tasks, where custom method has to be implemented, standard array methods are forbidden
Снимок экрана 2022-08-17 в 23 53 13

AlinaMatuschak (approved):
Okay, good job 👍

PR #938, opened 2022-08-13
done

adrqq:


yungdena (approved):
Good job! 👍

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let i = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      i = 1;
+    }
+
+    for (i; i < this.length; i++) {
+      prev = callback(prev, this[i], i, this);
+    }
```

AlinaMatuschak (changes requested):
Good 👍 , just use some array method

AlinaMatuschak (approved):
So, good job 👏

PR #940, opened 2022-08-14
Solution

vasyl-zinchenko:


zinchenko-anastasiia (approved):
nice job!!

src/arrayMethodReduce.js
```diff
@@ -3,9 +3,22 @@
 /**
  * Implement method Reduce
  */
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
dellachaise, line 9: I know that we use `prev` in video, but better rename to `previousValue` or `accumulator` as said in [method](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

dellachaise (approved):
Well done 👍

PR #941, opened 2022-08-14
Solution

VolodymyrChuyko:


AndreyJigijalkin (approved):
Good!)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,15 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const initialValueTest = !(arguments.length < 2);
```
```suggestion
    const haveInitialValue = !(arguments.length < 2);
```

AlinaMatuschak: Also, you can use just `!` operator 

```suggestion
    const initialValueTest = !!initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,15 @@
...
+    const startIndex = initialValueTest ? 0 : 1;
+    let accumulator = initialValueTest ? initialValue : this[0];
+
+    for (let i = startIndex; i < this.length; i++) {
```
AlinaMatuschak, line 12: We on `Array iteration methods` topic. So maybe should use some array method instead of a simple loop )

AlinaMatuschak (changes requested):
Good work 👏. Just small issues 👇

src/arrayMethodReduce.js
VolodymyrChuyko: done

VolodymyrChuyko: This won't work when undefined or zero is passed as an initialValue.

AlinaMatuschak: Too bad 🥲

AlinaMatuschak (approved):
🚀

PR #943, opened 2022-08-15
solution

Githi54:


KorniychukLesya (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let iteration = 0;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      iteration = 1;
+    }
+
+    for (iteration; iteration < this.length; iteration++) {
+      accumulator = callback(accumulator, this[iteration], iteration, this);
+    }
```

AlinaMatuschak (changes requested):
Awesome 😎, just use array method

AlinaMatuschak (approved):
So, good job 👏

PR #948, opened 2022-08-16
colutions

KorniychukLesya:


OlehMy (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
+    let result;
+    let start;
+
+    if (arguments.length === 2) {
+      result = initialValue;
+      start = 0;
+    } else {
+      result = this[0];
+      start = 1;
+    }
```
AlinaMatuschak, line 17: Then less code than better )

```suggestion
    let result = this[0];
    let start = 1;

    if (arguments.length === 2) {
      result = initialValue;
      start = 0;
    }
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
+
+    for (let i = start; i < this.length; i++) {
+      result = callback(result, this[i], i, this);
+    }
```
AlinaMatuschak, line 21: We on `Array iteration methods` topic. So maybe should use some array method instead of a simple loop )

AlinaMatuschak (changes requested):
Good 😊 , just use array method

AlinaMatuschak (approved):
So, good job 👏

PR #949, opened 2022-08-16
Solution

anyakrainiuk:


YuriiPronin (approved):

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, startValue) {
+    let prev = startValue;
+    let start = 0;
```
AlinaMatuschak, line 9:
```suggestion
    let accumulator = startValue;
    let startIndex = 0;
```

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
...
+
+    if (arguments.length < 2) {
+      start = 1;
+      prev = this[0];
+    }
+
+    for (let i = start; i < this.length; i++) {
+      prev = callback(prev, this[i], i, this);
+    }
```

AlinaMatuschak (changes requested):
Good 👍, just use array method

AlinaMatuschak (approved):
So, good job 👏

PR #954, opened 2022-08-17
Solution array method reduce

PastolNapas:


OleksiiPustovit (approved):
💪

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let sum = initialValue;
```
dellachaise: Better rename to `previousValue` or `accumulator` as said in [method](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let start = 0;
```
dellachaise: What does it mean `start` ? Your variable names should be meaningful. Better rename to `startIndex`.

dellachaise (changes requested):
Great job 👍

Just pay close attention to variable names 😉

dellachaise (approved):
Great job 🔥

PR #957, opened 2022-08-18
Done

andrii-dmytruk:


Mejeezz (approved):
Have no clue how am i supposed to reject this code. GJ!

src/arrayMethodReduce.js
```diff
@@ -5,8 +5,16 @@
...
+      ? callback(initialValue, this[0], 0, this) : this[0];
+
+    for (let i = 1; i < this.length; i++) {
+      prev = callback(prev, this[i], i, this);
+    }
+
+    return prev;
   };
 }
+applyCustomReduce();
```
```suggestion
```

src/arrayMethodReduce.js
```diff
@@ -5,8 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = arguments.length !== 1
+      ? callback(initialValue, this[0], 0, this) : this[0];
```
AlinaMatuschak: Try to call `[].reduce2((sum, num) => sum + num, 0);` with your reduce2 and reduce methods. The result will be different )

AlinaMatuschak (changes requested):
Almost perfect )

src/arrayMethodReduce.js
andrii-dmytruk: Ok, i don`t see the difference. But anyway i rewrote this to more readable logic

AlinaMatuschak: on your logic result will be NaN, but should be 0 )

AlinaMatuschak (approved):
🔥🔥

src/arrayMethodReduce.js
andrii-dmytruk: Oh, sorry. I didn`t test it with empty array. My bad

PR #959, opened 2022-08-18
solution

zinchenko-anastasiia:


Alzay007 (approved):
cool 👍  you can use only one loop , put a condition before the loop

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let currentValue = 0;
```
dellachaise: Better rename to `startIndex` because it is **index** not **value**.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
```
dellachaise, line 8: Great name 🔥

dellachaise (changes requested):
Great job 👍

Just pay close attention to variable names 😉

src/arrayMethodReduce.js
zinchenko-anastasiia: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce
I just used the terms of the documentation😅

danylo-onishchuk (approved):

PR #963, opened 2022-08-18
solution-reduce

NataliyaAn:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const start = (arguments.length < 2) ? 1 : 0;
```
dellachaise, line 8: Better rename to `startIndex`, because variable names should be meaningful.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
...
+    let step = (arguments.length < 2) ? this[0] : initialValue;
```
dellachaise, line 9: Better rename to `accumulator` for example, as in the original method.

dellachaise (approved):
Good job 👍 

Do not forget to ask for the student's reviews.

PR #974, opened 2022-08-24
Solution

mariabramnik:


stanislav-topikha (changes requested):
So for now your solution looks to complicated, try to get rid(or reduce) usage of `else`, `if else` and `if` nesting

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,27 @@
...
+    if (arguments.length === 1) {
+      if (typeof this[0] === 'number') {
+        startValue = this[0];
+        result = 0;
+      }
+    } else {
+      if (initialValue === undefined) {
+        result = 'undefined';
+      } else {
+        result = startValue = initialValue;
```
stanislav-topikha: it's not obvious how it works, make it simpler

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,27 @@
...
+    let startValue = '';
+
+    if (arguments.length === 1) {
+      if (typeof this[0] === 'number') {
+        startValue = this[0];
+        result = 0;
+      }
+    } else {
+      if (initialValue === undefined) {
+        result = 'undefined';
```
stanislav-topikha: if you need it, string (undefined)

stanislav-topikha (changes requested):
if you can try to reduce `if else` usage even more
comments ⬇️

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,24 @@
...
+    let startValue = (initialValue === undefined) ? '' : initialValue;
+
+    if (arguments.length === 1) {
+      if (typeof this[0] === 'number') {
+        startValue = this[0];
+        result = 0;
+      } else {
+        result = startValue = '';
+      }
+    }
```
liliia-livandovskaya:
```suggestion
    let startValue = initialValue;
    let startIndex = 0;

    if (arguments.length < 2) {
      startValue = this[0];
      startIndex = 1;
    }
```

liliia-livandovskaya (changes requested):
Try to rewrite your code to make it simplier

liliia-livandovskaya (approved):
🔥

PR #979, opened 2022-08-29
solution

vitalii-babiienko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = arguments.length < 2 ? this[0] : initialValue;
+    const startIndex = arguments.length < 2 ? 1 : 0;
```
stanislav-topikha: same condition, move it to the var

stanislav-topikha (changes requested):
comment ⬇️

stanislav-topikha (approved):

PR #997, opened 2022-09-16
Develop

grafit-off:
feat: method reduce was implemented

### DIFF
--- a/src/arrayMethodReduce.js
+++ b/src/arrayMethodReduce.js
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function (callback, initialValue) {
+    let startIndex = 0;
+    let prev = initialValue;
+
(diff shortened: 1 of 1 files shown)


varkovroma (approved):

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function (callback, initialValue) {
+    let startIndex = 0;
+    let prev = initialValue;
```
OleksiiBlezniuk, line 9:
```suggestion
    let acc = initialValue;
```

OleksiiBlezniuk: it's accumulator, not just prev value

OleksiiBlezniuk (approved):

src/arrayMethodReduce.js
grafit-off: Yes, agree with you. You write this one on the previous tasks, will change this soon

PR #998, opened 2022-09-17
reduce method

TheVovchik:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,22 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulated = initialValue;
```
iPanda177: as you said, variable name must start with noun, not a verb or adjective, right?

iPanda177 (approved):

src/arrayMethodReduce.js
TheVovchik: I think you are rigth here, will rename it 👍

src/arrayMethodReduce.js
```diff
@@ -5,8 +5,22 @@
...
+    let accumulator = initialValue;
+    let currentElementIndex = 0;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      currentElementIndex = 1;
+    }
+
+    for (let i = currentElementIndex; i < this.length; i++) {
+      const currentElement = this[i];
```
Anna-Tkachenko: this variable is redundant

Anna-Tkachenko (approved):

src/arrayMethodReduce.js
TheVovchik: it was added to better remember how it works, now it removed as useless

PR #1001, opened 2022-09-18
reduce created

illia-sosnytskyi:


Vladosik1337 (approved):
GJ)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let sum = initialValue;
```
silchyn: It is not necessary sum, precise name is accumulator for example

silchyn (changes requested):
Good job, but

silchyn (approved):
Cool!

PR #1011, opened 2022-09-20
solution

ihorandrianov:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,38 @@
...
-    // write code here
+    if (typeof callback !== 'function') {
+      throw new Error('not a function');
+    }
+
+    if (this.length === 0 && !initialValue) {
+      throw new Error('no initial value and array is empty');
+    }
+
+    let k = 0;
```
denyskleimenov, line 16: Try to give variables more precise names it will make the code more understandable.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,38 @@
...
+      : '';
+    let kPresent;
+
+    if (arguments.length === 2) {
+      accumulator = initialValue;
+    } else {
+      kPresent = false;
+
+      while (!kPresent && k < this.length) {
+        kPresent = this[k];
```
denyskleimenov, line 28: What's the point in using **kPresent**, could you explain? I'm just curious)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,38 @@
...
+    }
+
+    let k = 0;
+    let accumulator = typeof this[0] === 'number'
+      ? 0
+      : '';
+    let kPresent;
+
+    if (arguments.length === 2) {
+      accumulator = initialValue;
```
denyskleimenov, line 23: I think checking that we don't have enough arguments would be more appropriate because we expect to get at least 2 arguments for our function to work and if we have fewer, then we should do some actions. IMHO

denyskleimenov (approved):
Change variable names.

src/arrayMethodReduce.js
Andr1yK, line 16: What is `k`?

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,38 @@
...
+
+    if (this.length === 0 && !initialValue) {
+      throw new Error('no initial value and array is empty');
+    }
+
+    let k = 0;
+    let accumulator = typeof this[0] === 'number'
+      ? 0
+      : '';
+    let kPresent;
```
Andr1yK, line 20: what is `kPresent`?

Andr1yK (changes requested):
Fix naming :)

PR #1012, opened 2022-09-20
first working solution-it will be refactored ASAP

fatihsamur:


ghost (approved):
Approved, but I think mentors might not like the nested if's and loop in if statement

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let start = 0;
+
+    if (arguments.length <= 1) {
```
VinnieJ-2k20: Can use more obvious condition.
```suggestion
    if (arguments.length < 2) {
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #1017, opened 2022-09-21
Reduce method was implemented

denyskleimenov:


glu3s (approved):
🔥

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
Anna-Tkachenko, line 8: let's create more understandable variable name instead of `prev`

Anna-Tkachenko (approved):

PR #1020, opened 2022-09-21
method reduce was implemented

a-nahaitsev:


OleksandrGlulhovskiy (commented):
GJ!

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,23 @@
...
+  [].__proto__.reduce2 = function(callback, startValue) {
+    let currentValue;
+    let firstIndex;
+
+    if (arguments.length < 2) {
+      currentValue = this[0];
+      firstIndex = 1;
+    } else {
+      currentValue = startValue;
+      firstIndex = 0;
```
Anna-Tkachenko: let's set one of values by default for these variables

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #1034, opened 2022-09-22
solution

illia-kyselov:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let amount = initialValue;
+    let numOfIteration = 0;
```
SanyaBratashchuk:
```suggestion
    let startIndex = 0;
```

SanyaBratashchuk (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let amount = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      amount = this[0];
+      startIndex = 1;
+    }
```
Virass, line 15: Use ternary operator for set value for vars

Virass (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,24 @@
...
+=======
+    let amount = (arguments.length < 2) ? this[0] : initialValue;
+    const startIndex = (arguments.length < 2) ? 1 : 0;
```
loralevitska, line 18: In this case you don't need parentheses because you only have one condition

loralevitska (approved):

PR #1039, opened 2022-09-22
solution

Naveed-Shoukat:


fatihsamur (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,32 @@
...
+    if (arguments.length > 1 && initialValue === undefined) {
+      sum = 'undefined';
+      startIndex = 0;
+    }
+
+    // If initialValue is Provided
+    if (arguments.length > 1 && initialValue !== undefined) {
+      sum = initialValue;
+      startIndex = 0;
+    }
```
VinnieJ-2k20: This can be simplified using just one if.

VinnieJ-2k20 (changes requested):
Solution works, but can be simplified.

src/arrayMethodReduce.js
Naveed-Shoukat: required changes done

VinnieJ-2k20 (approved):

PR #1040, opened 2022-09-22
Implementation of reduce method

Dikey945:


glu3s (approved):
🔥

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = initialValue;
```
Anna-Tkachenko, line 8: let's create more understandable variable name instead of `acc`

Anna-Tkachenko (approved):

PR #1042, opened 2022-09-24
Solution

rstarovoit:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
grafit-off: it's accumulator, not just prev value

grafit-off (approved):
GJ 🔥

src/arrayMethodReduce.js
naziamoff: > it's accumulator, not just prev value

I agree here

naziamoff (approved):
Well done 🔥 
Just one small comment
- Due to `reduce` method logic it returns not the previous value but the accumulator. You should always take care of naming - it really matters

rstarovoit:
> Well done 🔥 Just one small comment
> 
> * Due to `reduce` method logic it returns not the previous value but the accumulator. You should always take care of naming - it really matters

Thanks! Fixed it!

naziamoff (approved):

PR #1055, opened 2022-10-02
Solution

emxm:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+    } else {
+      result = initialValue;
+
+      for (let i = 0; i < this.length; i++) {
+        result = callback(result, this[i], i, this);
+      }
+    }
+
+    return result;
```
SanyaBratashchuk: create an variable startIndex 0 by default and change its value if second argument is not passed to avoid duplication

SanyaBratashchuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1064, opened 2022-10-12
Solution for method Reduce

petrovskyi-yevhen:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
```
Anastasiia-Svintsova: You can use ternary operator here. For example
```suggestion
   const hasInitialValue = arguments.length < 2
   const prev = hasInitialValue ? this[0] : initialValue;
   const startIndex = hasInitialValue ? 1 : 0;
```

Anastasiia-Svintsova (changes requested):
Fix a comment, please

Bohdan-Maksymiuk (approved):

PR #1066, opened 2022-10-14
reduce applied

metinbicaksiz:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,23 @@
...
+      return initialValue;
+    }
+
+    let acc = initialValue;
+    let index = 0;
+
+    if (arguments.length < 2) {
+      acc = this[0];
+      index = 1;
+    }
```
amitpatiljc: can we do this with ternary operator?

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job 🔥

PR #1071, opened 2022-10-18
Solution

alexxxstep:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,37 @@
...
+      throw new Error('callback is not a function');
+    }
+
+    if (arguments.length >= 2) {
+      result = initialValue;
+    } else {
+      if (len === 0) {
+        throw new Error('reduce of empty array with no initial value');
+      }
+      result = obj[i];
```
Bohdan-Maksymiuk: you don't need all these custom errors here
simplify your solution
js will throw an error where it will be needed

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1108, opened 2022-11-02
Solution

alpopkov:


spiritkyiv (approved):
GJ!

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let previous = (arguments.length < 2) ? this[0] : initialValue;
+    const startIndex = (arguments.length < 2) ? 1 : 0;
```
denys-danyliuk, line 9: ternary expression should be multiline
```suggestion
    let previous = (arguments.length < 2) 
      ? this[0] 
      : initialValue;
    const startIndex = (arguments.length < 2) ? 1 : 0;
```

denys-danyliuk: also you can create one if statement where you will do this instead of creating two ternary expression

denys-danyliuk (changes requested):

PR #1109, opened 2022-11-02
add solution

AlexTsimba:


src/arrayMethodReduce.js
```diff
@@ -1,12 +1,27 @@
-'use strict';
+"use strict";
 
 /**
  * Implement method Reduce
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let total = initialValue;
```
vadymbaranov:
```suggestion
    let prevValue = initialValue;
```
total is quite a misleading name in this case.

vadymbaranov (approved):

src/arrayMethodReduce.js
denys-danyliuk: you can call it just 'index'

src/arrayMethodReduce.js
```diff
@@ -1,12 +1,27 @@
...
+    let startIndex = 0;
+
+    if (!this.length && !initialValue) {
+      throw new TypeError('Empty array and no initialValue');
+    }
```
denys-danyliuk: this is redundant

src/arrayMethodReduce.js
```diff
@@ -1,12 +1,27 @@
...
+
+    if (arguments.length < 2) {
+      total = this[0];
+      startIndex = 1;
+    }
+
+    for (let i = startIndex; i < this.length; i++) {
```
denys-danyliuk:
```suggestion
    for (startIndex; i < this.length; i++) {
```

denys-danyliuk (changes requested):

src/arrayMethodReduce.js
AlexTsimba: i tried to use this syntax but got "i is not defined" error so i tried to apply it in another way

illiarerun (approved):

PR #1111, opened 2022-11-02
complete

spiritkyiv:


andriy-shymkiv (commented):
why so many code 😅 . Have you seen @mgrinko videos about reduce method implementation?

you should use arguments, it is an array-like object that contains the arguments passed to the function.

```
function foo() {
  for (var i = 0; i < arguments.length; i++) {
    console.log(arguments[i]);
  }
}

foo(1,2,3);
//1
//2
//3
```

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,31 @@
...
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(...args) {
+    const callback = args[0];
+    let initialValue = args[1];
+
+    if (args.length === 2 && initialValue === undefined) {
+      initialValue = 'undefined';
+    }
```
vadymbaranov: I am lost in your variable naming, so used other names just which correspond to the variable functional meaning.
```suggestion
    if (args.length < 2) {
      previousValue = this[0];
      startIndex = 1;
    }
```

vadymbaranov (approved):
Interesting but complex solution. GJ!

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,31 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(...args) {
```
Andriy-Shymkiv, line 7: use `[].__proto__.reduce2 = function(arguments) {`

andriy-shymkiv: `[].__proto__.reduce2 = function(arguments) {`

andriy-shymkiv (changes requested):

G-YaSha (approved):

PR #1121, opened 2022-11-10
Reduce method solution

Sergii-Koviazin:


src/arrayMethodReduce.js
```diff
@@ -3,9 +3,22 @@
...
  * Implement method Reduce
  */
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
nazarmatsevych, line 12:
```suggestion
    if (initialValue === undefined) {
```

your version works well but is better if your condition strict

nazarmatsevych (approved):
Great 🔥

PR #1125, opened 2022-11-11
Solution

Irina-Kulish:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let res = initialValue;
```
nazarmatsevych, line 8: try not to cut variable names

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let start = 0;
+
+    if (arguments.length < 2) {
```
nazarmatsevych, line 11:
```suggestion
    if (initialValue === undefined) {
```

your version works well but is better if your condition strict

nazarmatsevych (approved):
Good work,
Check comments pls

PR #1137, opened 2022-11-25
solution

ysrckr:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
```
lefteris-L, line 12: You could declare i = 0 and change it to 1 in this if block. It would make your loop cleaner

lefteris-L (approved):
Great 🚀

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
...
+    }
+
+    for (let i = accumulator !== this[0] ? 0 : 1; i < this.length; i++) {
+      accumulator = callback(accumulator, this[i], i, this);
+    }
```
naziamoff: let's improve readability here

naziamoff (changes requested):

src/arrayMethodReduce.js
naziamoff: as I see now - you can create a variable for the condition you're trying to get here

naziamoff (approved):
🔥

PR #1140, opened 2022-11-26
solution

lefteris-L:


ysrckr (approved):
Looks good 🤟🏻

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = arguments.length < 2 ? this[0] : initialValue;
+    let i = arguments.length < 2 ? 1 : 0;
```
naziamoff, line 9: probably startIndex will sound better here?

naziamoff, line 9: ternaries should be multi-lined

naziamoff (approved):
Well done, but consider my comments pls

src/arrayMethodReduce.js
lefteris-L: I see, so, if `i` is declared outside the loop, it should have a clearer name, right?

PR #1141, opened 2022-11-27
Solution

nurdanokcu:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = arguments.length < 2 ? this[0] : initialValue;
+    const index = arguments.length < 2 ? 1 : 0;
```
ysrckr: You could set prev and index in a single if block. But everything looks good.

ysrckr (approved):

src/arrayMethodReduce.js
nurdanokcu: Thank you @ysrckr, I will consider your suggestion.

naziamoff: ternaries should be multi-lined, it improves readability

naziamoff (approved):

PR #1160, opened 2022-12-04
solution

vitalii-borzov:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      startIndex = 1;
+    }
```
tetiana-ivanchenko: it is good case to use ternary operator

tetiana-ivanchenko (changes requested):

vlad-demchuk (approved):

PR #1165, opened 2022-12-04
Solution

sercansimsek:


naziamoff (approved):
would be better to find a better name for `startsWith` variable. Basically it is indeed startsWith, although we return not startsWith but accumulator

PR #1171, opened 2022-12-05
solution

anastasiia-bilska:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let i = 0;
```
DiDemkiv: I think it's better to give the variable a more understandable name.
For example "startIndex"
```suggestion
    let startIndex = 0;
```

DiDemkiv (changes requested):
👍

DiDemkiv (approved):
Good job 🚀

solaryasha (approved):
👍

PR #1180, opened 2022-12-07
solution

Maks-Tsarenko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
```
tetiana-ivanchenko: it's a good case to use ternary operator to define the variable

tetiana-ivanchenko (changes requested):

src/arrayMethodReduce.js
stanislav-topikha: fix this 
image

stanislav-topikha (changes requested):

stanislav-topikha (approved):
GJ 👍🏻

PR #1187, opened 2022-12-12
first commit

haduigon:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let start = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      start = this[0];
+      startIndex = 1;
+    }
```
tetiana-ivanchenko: it's a goos case to use ternary operator for both variables f.e.:
const startIndex = arguments.length < 2 ? 1 : 0;

tetiana-ivanchenko (changes requested):

src/arrayMethodReduce.js
haduigon: done )

nazarmatsevych (approved):
👍

PR #1191, opened 2022-12-15
Solution

RomaSheva1987:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let start = 0;
```
Virass:
```suggestion
    let startIndex = 0;
```

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
...
+    let acc = initialValue;
```
Virass:
```suggestion
    let accumulator = initialValue;
```

Virass (approved):
Try to create names without abbreviations and, accordingly, from the logical content

AlisaVlasova (approved):

PR #1197, opened 2022-12-21
Solution

AnnyKoval:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const enoughArgs = arguments.length > 1;
+
+    const startIndex = enoughArgs ? 0 : 1;
+
+    let prev = enoughArgs ? initialValue : this[0];
```
Viktor-Kost: You use the same condition twice

We can simplify 
```
    let prev = initialValue
    let startIndex = 0 
    
    if (....) {
      prev = ...
      startIndex = ...
    }
```
```suggestion
```

Viktor-Kost (changes requested):
rejected

src/arrayMethodReduce.js
AnnyKoval: Thank you! Done✓

Viktor-Kost (approved):

PR #1206, opened 2022-12-27
initial commit

ihorivna007:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(accumulator, ...startValue) {
```
SanyaBratashchuk:
```suggestion
  [].__proto__.reduce2 = function(accumulator, initialValue) {
```

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
...
+    let result = this[0];
+    let start = 1;
+
+    if (startValue.length > 0) {
+      result = startValue[0];
+      start = 0;
+    }
```
SanyaBratashchuk:
```suggestion
    if (startValue.length >= 2) {
      result = initialValue;
      start = 0;
    }
```

SanyaBratashchuk (changes requested):

src/arrayMethodReduce.js
ihorivna007: Hello! Could you explain me this comment?
If we change '...startValue' to 'initialValue' at line 7, where will we get the 'startValue' for line 11?

src/arrayMethodReduce.js
ihorivna007: Why we can't use the ...args to handle undefined as a value?

loralevitska: Because the `initialValue` is one, there cannot be 3 or 4 of them. BTW: if you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
Fix your code by recommendation

src/arrayMethodReduce.js
loralevitska: The mentor suggests changes where appropriate. Further, if these changes concern other parts of the code, they must be changed manually. Also, if your code contains similar errors that the mentor pointed out, they also need to be fixed throughout your solution, not just in the comment area

Fix comment. Change `startValue` variable name to `initialValue` - all cases

loralevitska (changes requested):

ViktorHavryliuk (changes requested):
Feedback of previous mentors not implemented. Please fix issues highlighed by them. If you have questions, feel free to ask them in fe_chat1 and tag mentors if needed

OleziO (approved):

PR #1208, opened 2022-12-27
reduce

Vadym-Prydatok:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+    let startValue = 0;
+
+    if (arguments.length < 2) {
+      result = this[0];
+      startValue = 1;
+    }
```
tetiana-ivanchenko: it's perfect case to use ternary operator, f.e: 
`const startValue = arguments.length < 2 ? 1 : 0;`

tetiana-ivanchenko (changes requested):

ViktorHavryliuk (approved):

PR #1221, opened 2023-01-03
Solution

oleh-holovnykh:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator;
```
sv-dubov: Why is accumulator? :)

sv-dubov (approved):
Very interesting solution!

src/arrayMethodReduce.js
Anna-Tkachenko: it accumulates some values)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,30 @@
...
+        } else {
+          accumulator = this[0];
+          start = 1;
+        }
+        break;
+      default:
+        accumulator = initialValue;
+        start = 0;
+        break;
+    }
```
Anna-Tkachenko: good solution, but switch / case is redundant here
let's just simply set some initial values at the start of the function
and then according to the passed argument will change or won't change them

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #1228, opened 2023-01-05
Solution

anna-inozemets:


tania-boiar (approved):
Good job! 🔥

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,18 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = arguments.length < 2
```
Anna-Tkachenko, line 8: don't create shortened names for variables, will be better to use full naming

Anna-Tkachenko (approved):

PR #1232, opened 2023-01-05
reduce method was implemented

StetsenkoViktoriia:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
loralevitska:
```suggestion
    let prev = initialValue !== undefined ? initialValue : this[0];
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
...
+    let startIndex = 0;
+
+    if (initialValue === undefined) {
+      prev = 'undefined';
+    }
+
```
loralevitska:
```suggestion
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
```
loralevitska: I help you to improve `prev` value. Now improve `startIndex` by the example

loralevitska (changes requested):

src/arrayMethodReduce.js
StetsenkoViktoriia: in this way the test 'should handle undefined if it was passed as initialValue' doesn't work, so I have done a bit another way. or probably I didn't understand something?

loralevitska (approved):
😎

PR #1253, opened 2023-01-12
js / reduce method

valntyn:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
mykhalenych, line 8:
```suggestion
      let prev = arguments.length < 2 ? this[0] : initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let startIndex = 0;
```
mykhalenych, line 9: the same here instead if

mykhalenych (changes requested):

PR #1256, opened 2023-01-16
Solution

AV1788:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,37 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this == null) {
+      throw new Error('Can\'t iterate over undefined or null');
+    }
+
+    let result;
+    let i = 0;
+
+    const obj = Object(this);
```
vlad-demchuk, line 15: no need to move `this` to a separate variable

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,37 @@
...
+    const len = obj.length;
```
vlad-demchuk, line 16:
```suggestion
    const length = obj.length;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,37 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this == null) {
+      throw new Error('Can\'t iterate over undefined or null');
+    }
+
+    let result;
```
vlad-demchuk, line 12: specify variables at once after declaration

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,37 @@
...
+    let i = 0;
```
vlad-demchuk, line 13: use descriptive variable names

vlad-demchuk (changes requested):

PR #1260, opened 2023-01-19
Solution

novak-k:


src/arrayMethodReduce.js
```diff
@@ -3,9 +3,26 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length === 1) {
+      acc = this[0];
+      startIndex = 1;
+    } else if (arguments.length > 1 && initialValue === undefined) {
+      acc = 'undefined';
```
vlad-demchuk: why do you need `acc = 'undefined'`?

just check if arguments.length < 2  and do these actions
`acc = this[0];
startIndex = 1;`

vlad-demchuk (changes requested):

TarasHoliuk (approved):

PR #1267, opened 2023-01-25
Solution method Reduce

NazariiAlieksieiev:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
```
VitaliyBondarenko1982: Suggesting some refactor
```
const prev = initialValue || this[0];
const startIndex = initialValue ? 0 : 1;

vlad-demchuk (approved):

PR #1273, opened 2023-01-31
task was solved

vanyachyzh:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let value = arguments.length === 1 ? this[0] : initialValue;
```
VitaliyBondarenko1982, line 8:
```suggestion
    let value = initialValue || this[0]
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
...
+    const index = arguments.length === 1 ? 1 : 0;
```
VitaliyBondarenko1982, line 9:
```suggestion
    const index = initialValue ? 1 : 0;
```

VitaliyBondarenko1982 (approved):
Good job!
Just see my suggestions a,out little refactoring.

src/arrayMethodReduce.js
vanyachyzh: I check it and it doesn't work because we can put an empty string or 0 in the second argument. Both of them will be converted to boolean and it will be false. For example here we want to get string as a result but we will get a number [1,2,3,4].reduce((acc, e)=>{ return acc+ e }, "") because acc is false so we assign first element as initialValue

src/arrayMethodReduce.js
vanyachyzh: the same here

src/arrayMethodReduce.js
etojeDenys: yes, I agree, so we can make it better by just checking if initialValue != undefined
```suggestion
    let value = initialValue != undefined ? initialValue : this[0]
```

src/arrayMethodReduce.js
etojeDenys: same here
```suggestion
    const index = initialValue != undefined ? 0 : 1
```

etojeDenys (changes requested):

src/arrayMethodReduce.js
vanyachyzh: there is a case in test where we have undefined as a second argument and we have to use it as initialValue
image

VitaliyBondarenko1982: Yes, you are right.
checking arguments length looks better)
You do not need to re-request review again, because this task is already approved 🙂

PR #1280, opened 2023-02-03
function was created

YevheniiHura:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let previousValue = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      previousValue = this[0];
+      startIndex = 1;
+    }
```
nazarmatsevych, line 14: It's a good case to use ternary operator to set correct previosValue and startIndex

nazarmatsevych (approved):
Good work

PR #1294, opened 2023-02-09
solution

ivanm2706:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,26 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === 0) {
+        if (arguments.length === 1) {
+          accum = this[i];
+          i++;
+        } else {
+          accum = initialValue;
+        }
+      }
```
VitaliyBondarenko1982: You can avoid this nesting if else in loop  just make a couple of ternary expression before loop.
For accum, and for startIndex.
It wiil be more readable

VitaliyBondarenko1982 (changes requested):

IvanFesenko (approved):

PR #1296, opened 2023-02-10
Solution

TkachYuliia:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let start = 0;
```
etojeDenys:
```suggestion
    let startIndex = 0;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+
+    if (arguments.length < 2) {
```
etojeDenys, line 11:
```suggestion
    if (initialValue === undefined) {
```

etojeDenys (changes requested):

src/arrayMethodReduce.js
TkachYuliia: Can't use this recommendation because tests require code to handle undefined if it was passed as initialValue. Here you can see this test: https://prnt.sc/PgHvDcV4Fqgb

etojeDenys (approved):
LGTM

PR #1298, opened 2023-02-11
Solution

GUSILLUS:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
etojeDenys, line 11:
```suggestion
    if (initialValue === undefined) {
```

etojeDenys (changes requested):

src/arrayMethodReduce.js
GUSILLUS: If i do this i cant complete this test 'should handle undefined if it was passed as initialValue'
And this solution i done with lessons [video ](https://mate.academy/learn/javascript-advanced/array-iteration-methods#/video/146)

IvanFesenko (approved):

PR #1299, opened 2023-02-12
Solution

Ch0p3r84:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
```
Volodymyr-Melnyk-Fron10der, line 8: Just to be consistent with reduce naming
```suggestion
    let accumulator = initialValue;
```

Volodymyr-Melnyk-Fron10der (approved):
Perfect👍

PR #1300, opened 2023-02-13
reduce was implement

lilia-obushenko:


kuznetsova-anastasiia (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
OleksiiBlezniuk:
```suggestion
    let accumulator = initialValue;
```

OleksiiBlezniuk: It's not just `prev` - it's an `accumulator` for storing current reduce value

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
```
OleksiiBlezniuk: Or you could use ternary operator for assigning values to variables

OleksiiBlezniuk (approved):

PR #1303, opened 2023-02-14
solution

Spec0s:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let prev = initialValue;
```
Volodymyr-Melnyk-Fron10der: Make more meaningful names
```suggestion
    let value = initialValue;
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Almost perfect, just one remark here

Volodymyr-Melnyk-Fron10der (approved):
Great✅

PR #1311, opened 2023-02-18
Solution

Ssviatt:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
Volodymyr-Melnyk-Fron10der:
```suggestion
    let accumulator = initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let startIndex = 0;
+
+    if ((arguments.length) < 2) {
```
Volodymyr-Melnyk-Fron10der:
```suggestion
    if (arguments.length < 2) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great, just a small change I would do

ViktorHavryliuk (approved):

PR #1319, opened 2023-02-20
reduce method implementation

stanislavkulakovskyi:


timoxa2t (approved):
Good work

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
OleksiiBlezniuk, line 8:
```suggestion
    let accumulator = initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      startIndex = 1;
+    }
```
OleksiiBlezniuk, line 14: Or you could use ternary operator to assign values to variables

src/arrayMethodReduce.js
OleksiiBlezniuk: It's not just `prev` - it's an `accumulator` for storing current value

OleksiiBlezniuk (approved):

PR #1323, opened 2023-02-21
solution

Zelinskyi-Serhii:


stanislavcodes (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
```
OleksiiBlezniuk, line 8:
```suggestion
    let accumulator = initialValue;
```

OleksiiBlezniuk: It's not just a `result`, it's an `accumulator` for storing current reduce value

OleksiiBlezniuk (approved):

PR #1328, opened 2023-02-21
solution

milenadiamonds1:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
Yevhenii0536: maby it is a acc?

Yevhenii0536 (approved):

src/arrayMethodReduce.js
sergii-nosachenko: Why did you shorten the var name? It's not an optimization.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let startI = 0;
```
sergii-nosachenko: What is `I`?

sergii-nosachenko (changes requested):
Quick fixed are needed

sergii-nosachenko (approved):
Good work 👍

PR #1330, opened 2023-02-22
Solution

Mar4enkofff:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      startIndex = 1;
+    }
```
nazarmatsevych: It's a good case to use ternary operator for setting correct accumulator  and startIndex

nazarmatsevych (changes requested):

SanyaBratashchuk (approved):

PR #1332, opened 2023-02-22
solution

chernenko2:


YaroslavDzyadiv (approved):
Great job!

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
```
sergii-nosachenko: Cleanup

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
...
+    let prev = initialValue;
```
sergii-nosachenko: What is `prev`? Better name it as `accumulator`.

sergii-nosachenko (changes requested):
Quick fixes are needed

src/arrayMethodReduce.js
chernenko2: > What is `prev`? Better name it as `accumulator`.

Thank you for the advice, I changed it

sergii-nosachenko (approved):
Good work 👍

PR #1334, opened 2023-02-22
Solution for Array Method Reduce

ctk12:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = arguments.length < 2 ? this[0] : initialValue;
```
```suggestion
    let accumulator = arguments.length < 2 ? this[0] : initialValue;
```
It would be more clear to call it accumulator

polosanya (approved):
Good job!

PR #1337, opened 2023-02-22
Develop js_array-method-reduce

sergey-lipikhin:


VadKol (approved):
GJ

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let prev = initialValue;
```
OleksiiBlezniuk, line 9:
```suggestion
    let accumulator = initialValue;
```

OleksiiBlezniuk: It's not just a `prev` - it's an `accumulator` for accumulating reduce callback results

OleksiiBlezniuk (approved):

PR #1349, opened 2023-02-24
solution

NikitaStambul:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let callbackResult = initialValue;
+    let i = 0;
```
olecsiy-gavrilyuk: accumulator maybe will be better solution than callbackResult for name of variable and I think you can name variable i only in for loop in other situation would better name index

olecsiy-gavrilyuk (approved):

src/arrayMethodReduce.js
sergii-nosachenko: What is `i`?

sergii-nosachenko: Fix mate's comment

sergii-nosachenko (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let iteration = 0;
```
sergii-nosachenko, line 9:
```suggestion
    let index = 0;
```

sergii-nosachenko (approved):
Good work 👍

PR #1350, opened 2023-02-24
Array Method Reduce v.0.1 - Solution

NedobezhkinOleksandr:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,20 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let fromIndex = 0;
+
+    if (arguments.length < 2) {
+      fromIndex = 1;
+      prev = this[0];
+    }
```
nazarmatsevych: It's a good case to use ternary operator to set correct prev and fromIndex values

nazarmatsevych (changes requested):

NedobezhkinOleksandr:
Please, find updated code - from 'if' to 'ternary operator'

OleziO (approved):

PR #1352, opened 2023-02-24
My implementation of iteration array method "reduce"

gretoq:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if ([...arguments].length < 2) {
```
illiarerun, line 11: why do you use spread here?
```suggestion
    if (arguments.length < 2) {
```

illiarerun (approved):

PR #1368, opened 2023-03-03
custom reduce first try

RufatRM:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let startValue = initialValue;
```
Volodymyr-Melnyk-Fron10der: it's a better name 
```suggestion
    let accumulator = initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
...
+
+    if (arguments.length < 2) {
+      startValue = this[0];
+    }
+
+    const startIndex = (arguments.length < 2) ? 1 : 0;
```
Volodymyr-Melnyk-Fron10der: you can transform true into 1 and false into 0 right away
```suggestion
    const startIndex = Number(arguments.length < 2);
```

Volodymyr-Melnyk-Fron10der (approved):
Great. Just take a look on my small remarks to not repeat same mistakes on next time

PR #1373, opened 2023-03-07
:white_check_mark: implement custom reduce method

Bogdan-Mykhailov:
implement custom reduce method


sergii-nosachenko (approved):
Good work 👍

PR #1383, opened 2023-03-18
Solution

LBondarchuk:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      result = this[0];
+      startIndex = 1;
+    }
```
nazarmatsevych: It's a good case to use ternary operator to set correct result and startIndex

nazarmatsevych (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const startIndex = +(arguments.length === 1);
+    let result = startIndex ? this[0] : initialValue;
```
polosanya, line 9: It is better to avoid "result" and use more meaningful names)
```suggestion
    let accumulator = startIndex ? this[0] : initialValue;
```

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

PR #1385, opened 2023-03-20
Created reduce method

roshirin:


prokivskaaa (approved):
LGTM⭐️

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator;
+    let initValueFlag = 0;
+
+    if ('1' in arguments) {
```
SerhiiKirik, line 11: Interesting)
Also, you can check length)

SerhiiKirik (approved):
LGTM 🔥

package.json
roshirin: Got it. I've already made some more "git commit -am" pushes but will add only src files in the future.

PR #1389, opened 2023-03-22
reduce working

BondarenkoDmitriy:


kukla1989 (approved):
good job

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const hasInitialValue = arguments.length > 1;
+    let resultValue = hasInitialValue ? initialValue : this[0];
+
+    for (let i = hasInitialValue ? 0 : 1; i < this.length; i++) {
```
SerhiiKirik, line 11: It would be cleaner 
```suggestion
    const hasInitialValue = arguments.length > 1;
    const startIndex = hasInitialValue ? 0 : 1;
    let resultValue = hasInitialValue ? initialValue : this[0];

    for (let index = startIndex; index < this.length; index++) {
```

SerhiiKirik (approved):
Great 🔥

PR #1393, opened 2023-03-22
solution

Sbavdik:


Stambuliak (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const hasInitialValue = arguments.length > 1;
+    let sum = arguments.length > 1 ? initialValue : this[0];
+
+    for (let i = hasInitialValue ? 0 : 1; i < this.length; i++) {
```
SerhiiKirik: It would be cleaner
Also, you have hasInitialValue, don't remember about this)
```suggestion
    const hasInitialValue = arguments.length > 1;
    const startIndex = hasInitialValue ? 0 : 1;
    let sum = hasInitialValue ? initialValue : this[0];

    for (let index = startIndex; index < this.length; index++) {
```

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
Nice 🔥

PR #1396, opened 2023-03-22
solution custom reduce method

didyk-ma:


vinnykalisa (approved):
Good job! I'm not sure about the sum being a semantically valid variable name in this case.

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let sum = initialValue;
```
SerhiiKirik, line 8:
```suggestion
    let accumulator = initialValue;
```

SerhiiKirik (approved):
Well done 🚀

PR #1414, opened 2023-03-25
Fedenko-Method-Reduce

VladFedenko:


etojeDenys (changes requested):
looks good,

but we also need to check if the initial index equals 0 or 1

anstsot (approved):
Great job, approved!

PR #1416, opened 2023-03-26
create function custom reduce

VenomMB:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let suma = initialValue;
```
SerhiiKirik, line 8:
```suggestion
    let accumulator = initialValue;
```

SerhiiKirik (approved):
Well done 🚀

PR #1421, opened 2023-03-28
1st commit

ihor-halkin:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initialValue = 'undefined') {
```
sTorba24:
```suggestion
  [].__proto__.reduce2 = function(callback, initialValue) {
```

sTorba24: not passed argument is undefined by default as not declared variable

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
...
+    let prevValue = initialValue;
```
sTorba24:
```suggestion
    let accumulator = initialValue;
```

sTorba24: in official documentation it is accumulator, it's not prevValue, it's value that store result of sum of callbacks for prev values

sTorba24 (changes requested):
Good work! lets improve a little

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if (initialValue === undefined) {
+      accumulator = 'undefined';
+    }
```
sTorba24: you don't need it, if you didn't pass initialValue it's undefined already, you didn't define it so => undefined

sTorba24 (changes requested):
Little thing to fix

src/arrayMethodReduce.js
ihor-halkin: got it

witflash (approved):
Well done 👍🏻

PR #1423, opened 2023-03-28
done

victor-buldenko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let start = this[0];
+    let i = 1;
```
mykhalenych:
```suggestion
    let startIndex = 1;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+
+    if (arguments.length === 2) {
+      start = initialValue;
+      i = 0;
+    }
+
+    for (; i < this.length; i++) {
```
mykhalenych:
```suggestion
    for (let i  = startIndex; i < this.length; i++) {
```

mykhalenych (changes requested):

src/arrayMethodReduce.js
victor-buldenko: you are maybe joke?

victor-buldenko: i have iterator `i`, why i must create second variable for doing same?

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
-    // write code here
+    let start = this[0];
+    let sameShitAsIndex = 1;
+
+    if (arguments.length === 2) {
+      start = initialValue;
+      sameShitAsIndex = 0;
+    }
+
+    for (let index = sameShitAsIndex; index < this.length; index++) {
```
loralevitska: for better understanding of the code, it is a "best practice" to declare the index in the body of the loop. We improve the code style so that our colleagues can understand it and quickly understand other people's (our) code

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let start = this[0];
+    let sameShitAsIndex = 1;
```
loralevitska: Bad variable name
Rename the variable according to the rules for naming variables
image
ex `startIndex` `commonIndex` `innerIndex` `index` (and in loop will be let i = index)

loralevitska (changes requested):
Almost done) Let's improve variable naming
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

loralevitska (approved):

PR #1440, opened 2023-04-06
solution

MikiOOOO:


ghostargg (approved):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
...
+
+    let currentValue;
+    let startingIndex = 0;
+
+    if (arguments.length === 2) {
+      currentValue = initialValue;
+    } else {
+      currentValue = this[0];
+      startingIndex++;
+    }
```
d-wasilewski: maybe just check if initialValue exists. If it does, assign it to a variable, otherwise assign this[0]. You can use ternary and do it in 1 line

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1448, opened 2023-04-10
Solution

MaksymKos:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let firstValue = initialValue;
+    let i = 0;
+
+    if (!(1 in arguments)) {
+      firstValue = this[0];
+      i++;
+    }
+
```
IvanFesenko, line 15: Better will be set index instead of post increment
```
    if (arguments.length < 2) {
      firstValue = this[0];
      i = 1;
    }
```

IvanFesenko (approved):

PR #1450, opened 2023-04-10
first version

Vladimir-Zadorozhnyi:


tetiana-ivanchenko (changes requested):
you didn't use second argument of function

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,13 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = arguments.length > 1 ? initialValue : this[0];
```
polosanya: I think it's better to check if initialValue is not undefined

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,13 @@
...
+
+    for (let i = accumulator === this[0] ? 1 : 0; i < this.length; i++) {
```
polosanya: Same here. What if `this[0]` will accidentally be the same as `initialValue`? You will skip one iteration in that case

polosanya (changes requested):
Cool solution! Just found one small moment that would be great to fix)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator;
+    let startIndex;
```
polosanya: You can declare them with correct values and just change them if arguments.length is > 2. This will help to avoid extra checks

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,28 @@
...
+
+    if (arguments.length > 1) {
+      if (initialValue === undefined && this.length === 0) {
+        throw new TypeError('Reduce of empty array with no initial value');
```
polosanya: Actually, there shouldn't be any errors. The original reduce method works with an empty array and no initial value

polosanya (changes requested):
Almost done!

OleziO (approved):

PR #1452, opened 2023-04-11
Solution

RomanOstrous:


src/arrayMethodReduce.js
```diff
@@ -1,12 +1,18 @@
...
+'use strict';
+
+/**
+ * Implement method Reduce
+ */
+function applyCustomReduce() {
+  [].__proto__.reduce2 = function(callback, initialValue) {
+    let acc = arguments.length === 1 ? this[0] : initialValue;
+
+    for (let i = arguments.length === 1 ? 1 : 0; i < this.length; i++) {
```
witflash, line 10: It's better to move `startIndex` to the separate constant. It should looks better

witflash (approved):
Good work! 👍🏻

PR #1473, opened 2023-04-27
Solution

ollavka:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prevValue = initialValue;
+    let startIdx = 0;
+
+    if (arguments.length === 1) {
+      prevValue = this[0];
+      startIdx = 1;
+    }
```
tetiana-ivanchenko, line 14: it's good case to use ternary operator

tetiana-ivanchenko (approved):

PR #1476, opened 2023-04-30
Reduce solution

IzzetYusufov:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
+    let start = 0;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      start = 1;
+    }
```
nazarmatsevych, line 14: It's a good case to use ternary to set correct prev and start values

nazarmatsevych (approved):

PR #1477, opened 2023-05-01
Implement_method_Reduce

rialleons:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startingIndex = 0;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      startingIndex++;
+    }
```
tetiana-ivanchenko: good case to use ternary operator

tetiana-ivanchenko (approved):

src/arrayMethodReduce.js
rialleons: Like this:
```suggestion
    let accumulator = arguments.length < 2
      ? this[0]
      : initialValue;
    const startingIndex = arguments.length < 2 ? 1 : 0;
    }
```
?

PR #1479, opened 2023-05-02
1.0.0 10 of 11 passed

domkab:
i dont understand check for undefined, please help

### DIFF
--- a/src/arrayMethodReduce.js
+++ b/src/arrayMethodReduce.js
@@ -1,11 +1,15 @@
 'use strict';
 
-/**
- * Implement method Reduce
- */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue !== undefined ? initialValue : this[0];
+    const start = initialValue !== undefined ? 0 : 1;
+
+    for (let i = start; i < this.length; i++) {
+      accumulator = callback(accumulator, this[i], i, this);
+    }
+
+    return accumulator;
   };
 }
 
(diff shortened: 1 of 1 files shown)


src/arrayMethodReduce.js
```diff
@@ -1,11 +1,15 @@
 'use strict';
 
-/**
- * Implement method Reduce
- */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue !== undefined ? initialValue : this[0];
+    const start = initialValue !== undefined ? 0 : 1;
```
qoa11a: In this case, if the user passes `undefined` as an `initialValue` it won't work correctly.
You could check if the `initialValue` was passed in this way: `if (arguments.length < 2) { ... }`.

qoa11a (changes requested):
Here's a hint. ;)

src/arrayMethodReduce.js
```diff
@@ -1,11 +1,23 @@
...
+    let accumulator;
+    let start;
+
+    if (arguments.length >= 2) {
+      accumulator = initialValue;
+      start = 0;
+    } else {
+      accumulator = this[0];
+      start = 1;
+    }
```
Anastasiia-Svintsova, line 14: You can shorten it
```suggestion
    let accumulator = this[0];
    let start = 1;

    if (arguments.length >= 2) {
      accumulator = initialValue;
      start = 0;
    }
```

Anastasiia-Svintsova (approved):

PR #1480, opened 2023-05-03
solution

generatedman:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
ghost, line 8: yea, we can also use `prev` as the variable name, but it would be more correct to call it `accumulator`
```suggestion
    let accumulator = initialValue;
```

ghost (approved):

33Orange (approved):
Well done!

PR #1484, opened 2023-05-03
Add solution

ghost:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
33Orange, line 11: Use a default parameter for 'initialValue' instead of checking the length of the 'arguments' object. For example: 'function(callback, initialValue = this[0])'.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+      accumulator = this[0];
```
33Orange, line 12: Replace 'arguments.length' with a check for 'initialValue === undefined' to determine whether an initial value was provided.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
```
33Orange, line 7: Avoid modifying the prototype of built-in objects. Instead, create a standalone function that takes the array as an argument along with the callback and initialValue.

33Orange (approved):
Well done!

PR #1487, opened 2023-05-04
 js_array-method-reduce

Donovanoff:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let currentIndex = 0;
+    let accumulator = initialValue;
+
+    if (arguments.length < 2) {
+      currentIndex = 1;
+      accumulator = this[0];
+    }
```
tetiana-ivanchenko, line 14: good case to use ternary operator

tetiana-ivanchenko (approved):

PR #1488, opened 2023-05-06
Solution

vintwp:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator;
```
volodymyr-soltys97: Let's write down the initial value here
```suggestion
    let accumulator = initialValue;
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
...
+    let initialIndex = 0;
+
+    if (arguments.length === 1) {
+      accumulator = this[0];
+      initialIndex = 1;
+    } else if (arguments.length === 2 && !initialValue && initialValue !== 0) {
+      accumulator = String(initialValue);
+    } else {
+      accumulator = initialValue;
+    }
```
volodymyr-soltys97: Try to use only one construction if instead if, else if, else.
You can check if the length is less than 2

volodymyr-soltys97 (changes requested):
Good job. 👍 
But let's make your code more readable

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let initialIndex = 0;
+
+    if (arguments.length < 2 && initialValue !== 0) {
```
Volodymyr-Melnyk-Fron10der: if there is only 1 argument then initialValue will be undefined anyway, so I would remove the `&& initialValue !== 0`

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+      accumulator = this[0];
```
Volodymyr-Melnyk-Fron10der: It is better to use `this[0]` only when the length of the array is greater than 0, to avoid accessing undefined elements.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+      initialIndex = 1;
```
Volodymyr-Melnyk-Fron10der: It is better to use `this[0]` only when the length of the array is greater than 0, to avoid accessing undefined elements.

Volodymyr-Melnyk-Fron10der (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let initialIndex = 0;
+
+    if (arguments.length < 2 && this.length > 1) {
+      accumulator = this[0];
+      initialIndex++;
```
nazarmatsevych:
```suggestion
      initialIndex = 1;
```

i think this way is more obvious

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    }
```
nazarmatsevych: It's a good case to use ternary to set correct accumulator  and initialIndex

nazarmatsevych (approved):

PR #1494, opened 2023-05-09
add reduceCopy solution

DidenkoDenis306:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = arguments.length > 1 ? initialValue : this[0];
```
ghost: let's make our variables clearer and call it `accumulator` instead of `acc`

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
...
+    const start = arguments.length > 1 ? 0 : 1;
```
ghost: do we need such complicated checks? we make the same thing in one `if` statement

ghost (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let start = 0;
```
dinaminahi: Can we come up with more descriptive name here? something like `startIndex`

dinaminahi (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1499, opened 2023-05-11
task solved

zisuan-od:


src/arrayMethodReduce.js
```diff
@@ -3,9 +3,25 @@
...
-    // write code here
+    let accumulator;
+    let startLoop;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      startLoop = 1;
+    } else {
+      accumulator = initialValue;
+      startLoop = 0;
```
DarkMistyRoom, line 17: you can avoid `else `statement by setting these values when declaring

DarkMistyRoom (approved):
Good job

PR #1506, opened 2023-05-13
Solution

yuliia-nudyk:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,23 @@
...
+    let prev;
+    let fromIndex;
+
+    if (arguments.length < 2) {
+      prev = this[0];
+      fromIndex = 1;
+    } else {
+      prev = initialValue;
+      fromIndex = 0;
+    }
```
agniya-i, line 17:
```suggestion
    let prev = initialValue;
    let fromIndex = 0;

    if (arguments.length < 2) {
      prev = this[0];
      fromIndex = 1;
    } 
```

You can also set these values by default and remove the else block to make code more simpler

agniya-i (approved):

PR #1518, opened 2023-05-18
Solution

julshvets:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,24 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prevValue = initialValue;
```
33Orange: Would be better to name it `accumulator`, cause it's not just past value

33Orange (changes requested):
Well done !

Just a one moment with naming.

dinaminahi (approved):
Good job! 👍

PR #1520, opened 2023-05-18
Solution

ychaiun:


src/arrayMethodReduce.js
```diff
@@ -3,10 +3,23 @@
 /**
  * Implement method Reduce
  */
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let startIndex = 0;
+    let accum = initialValue;
```
polosanya, line 10:
```suggestion
    let accumulator = initialValue;
```
It's not critical, but better to use full names

polosanya (approved):
Nice!

PR #1521, opened 2023-05-18
Solution

ivankhrsh:


33Orange (approved):
Well done ! 🔥

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let previous = initialValue;
```
33Orange, line 8: It is better to use the name `accumulator`. This is not the previous value

33Orange (approved):

PR #1526, opened 2023-05-19
reduce method add solution

YeMyr:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
```
DidenkoDenis306: You dont need to use the spread operator (...initialValue) in the function parameter list, as it was unnecessary. The initialValue should be a single value, not an array.

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
...
+    const prev = initialValue;
```
DidenkoDenis306: Woulde be better to rename this variable to "accumulator"

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
...
+
+    for (let i = 0; i < this.length; i++) {
```
DidenkoDenis306: create a startIndex variable with default value 0. and use it instead i = 0. Move your if statment before the loop.

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
...
+      if (prev.length === 0) {
```
DidenkoDenis306: you should check if arguments.length < 2

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
...
+        prev[0] = this[i];
+        i++;
+      }
+
+      prev[0] = callback(prev[0], this[i], i, this);
```
DidenkoDenis306: use accumulator instead of element of prev.

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,19 @@
...
+    }
+
+    return prev[0];
```
DidenkoDenis306: the same, use accumulator.

DidenkoDenis306 (changes requested):
Good job, a few moments to fix)

src/arrayMethodReduce.js
YeMyr: 👌 fixed

src/arrayMethodReduce.js
YeMyr: 👌 fixed

src/arrayMethodReduce.js
YeMyr: 👌 fixed

src/arrayMethodReduce.js
YeMyr: 👌 fixed

src/arrayMethodReduce.js
YeMyr: 👌 fixed

src/arrayMethodReduce.js
YeMyr: 👌 fixed

YeMyr:
@DidenkoDenis306 Hi! Thanks for review! I've revised mistakes, please check, if there is everything ok.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startValue = 0;
```
DidenkoDenis306: Rename this variable to startIndex, because this is the index of element of the array from which we will start in the loop

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+
+    if ([...arguments].length < 2) {
+      accumulator = this[0];
+      startValue++;
```
DidenkoDenis306: It would be better to assign a value (startIndex = 1) instead of just adding.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startValue = 0;
+
+    if ([...arguments].length < 2) {
```
DidenkoDenis306: You can change it to arguments.length without spread operator

DidenkoDenis306 (changes requested):
Super!!! There are a couple of little mistakes left to correct.

src/arrayMethodReduce.js
YeMyr: done 👌
oh, yes, exactly! for some reason I like to spread everything))))

src/arrayMethodReduce.js
YeMyr: done 👌

src/arrayMethodReduce.js
YeMyr: Done 👌

DidenkoDenis306 (approved):
🚀🚀🚀

silchyn (approved):

PR #1527, opened 2023-05-19
reduce method

1v43k:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, initalValue, index = 0) {
```
etojeDenys: The 'index' parameter is unnecessary. Remove it and update the function accordingly.

etojeDenys (changes requested):

mykhalenych (approved):

PR #1528, opened 2023-05-20
add task solution

KatyaZaremska:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let previous = initialValue;
```
33Orange: Use more descriptive names.
It's not `previous` value. It's accumulator which have sum of all prev actions

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let start = 0;
```
33Orange:
```suggestion
    let initialIndex = 0;
```
Would be better

33Orange (changes requested):
Well done ! 🔥 

A few moments with readability.

silchyn (approved):

PR #1530, opened 2023-05-20
reduce

diana-tuz:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let startValue = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length === 1) {
+      startValue = this[0];
+      startIndex = 1;
+    }
```
nazarmatsevych, line 14: It's a good case to use ternary to set correct startValue and startIndex

nazarmatsevych (approved):
👍

PR #1539, opened 2023-05-24
code

liza-strykharchuk:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let preValue = initialValue;
+
+    if (arguments.length < 2) {
```
volodymyr-soltys97, line 11: Instead of checking 'arguments.length', consider checking if 'initialValue' is undefined.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+      preValue = this[0];
+
+      for (let i = 1; i < this.length; i++) {
```
volodymyr-soltys97: Consider using a single for loop with a conditional to handle the case when 'initialValue' is undefined.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/arrayMethodReduce.js
liza-strykharchuk: if i change this statmet my code will return error when undefined was passed as initialValue

volodymyr-soltys97 (approved):
Great 🔥

PR #1550, opened 2023-05-31
Solution

ArtemVelychko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this.length === 0) {
+      return initialValue !== undefined ? initialValue : 'undefined';
```
Anastasiia-Svintsova: Returning 'undefined' as a string is not correct. You just can add a default value to the initialValue if you really need it
You can just remove this if statement, it's redundant

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
...
+    }
+
+    let startIdx = 0;
+    let result;
+
+    if (arguments.length === 2) {
+      result = initialValue;
+    } else {
```
Anastasiia-Svintsova: Let's simplify it a little bit
```suggestion
    let result = initialValue;

    if (... (add the correct condition here)) {
```

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):

PR #1554, opened 2023-06-01
Solution

IraPolchanin:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let prev = initialValue;
```
```suggestion
    let accumulator = initialValue;
```
It would be more accurate to call it 'accumulator'

polosanya (approved):
Great!

PR #1560, opened 2023-06-08
solution

EvgenGolub:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let startValue = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
volodymyr-soltys97, line 11: Instead of checking the arguments length, you can use 'initialValue === undefined'

volodymyr-soltys97 (approved):
Nice work 👍

PR #1564, opened 2023-06-11
arrayMethodReduce v1

Voloshyn24:


src/arrayMethodReduce.js
```diff
@@ -6,6 +6,19 @@
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
     // write code here
+    let startIndex = 0;
+    let result = initialValue;
```
polosanya, line 10:
```suggestion
    let accumulator = initialValue;
```
It would be great to use more meaningful name than just `result`

polosanya (approved):
Well done!

PR #1565, opened 2023-06-11
Solution MR

RuslanMelnykFE:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const args = [...arguments];
```
anastasiiavorobiova, line 8: Instead of using 'const args = [...arguments];', consider using the function signature 'function(callback, initialValue)'.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
+    let startIndex = 0;
+    let result = null;
+
+    if (args.length < 2) {
```
anastasiiavorobiova, line 12: Replace 'if (args.length < 2)' with 'if (initialValue === undefined)'. This will handle the case when initialValue is not provided.

anastasiiavorobiova (changes requested):
Great job!
Needs tiny refactoring

src/arrayMethodReduce.js
RuslanMelnykFE: Використати таку умову (initialValue === undefined) не вийде так як не спрацює тест

test('should handle undefined if it was passed as initialValue', () => {
  expect(sourceStr.reduce2((acc, current) => acc + current, undefined))
    .toBe('undefinedabcd');
});

І також через це використовув [...arguments]

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const args = [...arguments];
+    let startIndex = 0;
+    let result = null;
```
volodymyr-soltys97, line 10: It's better to initialize 'result' with 'initialValue' to avoid potential issues when the initialValue is not provided.

volodymyr-soltys97 (approved):
Nice work 👍

PR #1574, opened 2023-06-15
task solution

Bonum-taurus:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let index = 0;
+
+    if (arguments.length < 2) {
```
volodymyr-soltys97, line 11: Also you can replace 'arguments.length' with '(typeof initialValue !== 'undefined')' to improve readability and check if initialValue is provided.

volodymyr-soltys97 (approved):
Great job 👍

PR #1577, opened 2023-06-16
Solution

adv555:


cebanoleksandr (approved):
Nice job:)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,18 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = initialValue;
+
+    for (let i = 0; i < this.length; i++) {
+      if (acc === undefined && arguments.length === 1) {
```
juljaVysotska, line 11: In you have only 1 argument, your second argument will be always undefined. So, are we actually need check for `acc` here?

juljaVysotska (approved):
Good job!

PR #1581, opened 2023-06-17
Solution

rvrazovskyi:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,33 @@
...
+    if (arguments.length >= 2) {
+      accumulator = initialValue;
+      startIndex = 0;
+    } else {
+      accumulator = array[0];
+      startIndex = 1;
+    }
+
+    for (let i = startIndex; i < length; i++) {
+      accumulator = callback(accumulator, array[i], i, array);
```
irini132, line 31: Consider using the `array-callback-return` ESLint rule to ensure that the callback function in `reduce2` always has a return statement.

irini132 (commented):
Check the comment

PR #1587, opened 2023-06-19
add solution

maxyma-sanx:


IraShtynda (approved):
Good job👍

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = arguments.length > 1 ? initialValue : this[0];
+    const startIndex = arguments.length > 1 ? 0 : 1;
+
+    for (let i = startIndex; i < this.length; i += 1) {
+      if ((this[i], i, this)) {
```
ManoylovAC: This is a very unclear condition - please rewrite it in a more understandable way and explain the purpose if the condition is necessary

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc = arguments.length > 1 ? initialValue : this[0];
```
ManoylovAC: It will be better to use more clear name instead of `acc`.

ManoylovAC (changes requested):
Good start
Please check my comments I left and make improvements

ManoylovAC (approved):
Good 👍

PR #1592, opened 2023-06-19
Solution

dianadiv:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,28 @@
...
+
+    if (arguments.length === 2) {
+      accumulator = initialValue;
+
+      findResult(0, this);
+    } else {
+      accumulator = this[0];
+
+      findResult(1, this);
+    }
```
outofs: Good work, but I prefer using ternary operator instead of if/else statement. You could just make:
```
let accumulator = arguments.length === 2? initialValue : this[0];
const index = arguments.length === 2? 0 : 1;
```
And with that you don't need separate `findResult ` function .

outofs (approved):
Well done! Approve!

src/arrayMethodReduce.js
```diff
@@ -5,8 +5,28 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+
+    let accumulator;
```
YegorKorenkov:
```suggestion
    let accumulator = this[0];
    let index = 1;
```

src/arrayMethodReduce.js
YegorKorenkov:
```suggestion
    if (arguments.length === 2) {
      accumulator = initialValue;
      index = 0;
    }
```

YegorKorenkov: Good tip! You can add additional **startIndex** variable and use it in for loop without adding extra function **findResult**. This solution

YegorKorenkov (changes requested):

anstsot (approved):
Good work, approving:sparkles:

PR #1599, opened 2023-06-20
add task solution

irynabolbat:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
...
+    let reducedValue = initialValue;
+    let index = 0;
+
+    reducedValue = arguments.length < 2
+      ? this[0]
+      : reducedValue;
+
+    index = arguments.length < 2
+      ? 1
+      : index;
```
MarianVolytskyi: you don't need to overwrite variables, you can make checking right there
```
let reducedValue = arguments.length < 2
      ? this[0]
      : initialValue;

  let index = arguments.length < 2
      ? 1
      : 0;
```

MarianVolytskyi (approved):
Good job!
Please read my comments.

src/arrayMethodReduce.js
tetiana-ivanchenko:
```suggestion
    const reducedValue = arguments.length < 2
      ? this[0]
      : initialValue;
    const index = arguments.length < 2
      ? 1
      : 0;
```

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1603, opened 2023-06-21
reduce method solution

ivanbilotserkivskiy:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,21 @@
...
-    // write code here
+  [].__proto__.reduce2 = function(...args) {
+    const [callback, initialValue] = args;
+    let startIndex = 0;
+    let result = initialValue;
+
+    if (args.length === 1) {
+      startIndex = 1;
+      result = this[0];
+    }
```
NataliGru, line 15: You could use `arguments` to access the array of arguments instead of using destructuring. Additionally, you can use a ternary operator to set the values of `startIndex` and `result` variables✨

NataliGru (approved):
Good job!) Please pay attention to the comment below✨

tetiana-ivanchenko (approved):

PR #1604, opened 2023-06-21
arrayMethodReduce Solution

NataliGru:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = arguments.length > 1 ? initialValue : this[0];
+    const startIndex = arguments.length > 1 ? 0 : 1;
+
+    for (let i = startIndex; i < this.length; i++) {
+      const currentValue = this[i];
+      const currentIndex = i;
```
vladmakohon, line 13: These variables just seem unnecessary.

vladmakohon (approved):
Good solution!

src/arrayMethodReduce.js
NataliGru: Yeah, I agree with you in this case))✨
I rewrote it as an example on MDN, just for fun) 
I really like creating variables for a clearer code style in the future. It's a good practice)

ManoylovAC (approved):
Clear and clean 👍

PR #1610, opened 2023-06-24
solution

Pa1eOrc:


src/arrayMethodReduce.js
```diff
@@ -1,11 +1,28 @@
...
- */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc;
+
+    if (initialValue !== undefined) {
+      acc = initialValue;
+    } else if (arguments.length > 1) {
+      acc = initialValue;
```
OleziO: Remove this `else if`

src/arrayMethodReduce.js
```diff
@@ -1,11 +1,28 @@
...
+    } else {
+      acc = this[0];
+    }
```
OleziO: Replace with ternary

src/arrayMethodReduce.js
```diff
@@ -1,11 +1,28 @@
...
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acc;
+
+    if (initialValue !== undefined) {
+      acc = initialValue;
+    } else if (arguments.length > 1) {
+      acc = initialValue;
+    } else {
+      acc = this[0];
```
OleziO: Replace with `acc = 0`

src/arrayMethodReduce.js
```diff
@@ -1,11 +1,28 @@
...
+    }
+
+    if (acc === this[0]) {
+      for (let i = 1; i < this.length; i++) {
+        acc = callback(acc, this[i], i, this);
+      }
+    } else {
+      for (let i = 0; i < this.length; i++) {
+        acc = callback(acc, this[i], i, this);
```
OleziO: Leave only this cycle, the other one has been issued. The condition is not needed because you changed `acc = this[0]` to `acc = 0`

OleziO (changes requested):

src/arrayMethodReduce.js
Pa1eOrc: Код з цими виправленнями не проходить тести

src/arrayMethodReduce.js
Pa1eOrc: Код з цими виправленнями не проходить тести

src/arrayMethodReduce.js
```diff
@@ -1,12 +1,50 @@
...
+      }
+    }
+
+    return acc;
   };
 }
 
 module.exports = applyCustomReduce;
+
+// function applyCustomReduce() {
```
loralevitska: Remove comments. Try it write obvious code that explain all we need without comments)

src/arrayMethodReduce.js
loralevitska: Accepted)

loralevitska: But you have similar code, so remove redundant repeats and change conditionally  only index value 
image

loralevitska (changes requested):
Almost done) Let's improve

OleziO (approved):

PR #1611, opened 2023-06-24
reduce

alina-isakova:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,14 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accum = arguments.length === 2 ? initialValue : this[0];
```
OleziO: Move `arguments.length === 2` to the variable (It optimizes your app, because the condition will be checked only 1 time, not 2 times)

OleziO (changes requested):

src/arrayMethodReduce.js
OleziO: Not resolved.

```
const checkArgLength = arguments.length === 2;
let accum = checkArgLength ? initialValue : this[0];
const start = checkArgLength ? 0 : 1;
```

OleziO (changes requested):

src/arrayMethodReduce.js
alina-isakova: thank you, done

etojeDenys (approved):
lgtm

PR #1613, opened 2023-06-28
Solution

yurii-krymskiy:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let resultValue = initialValue;
+    let startValue = 0;
+
+    if (arguments.length < 2) {
+      resultValue = this[0];
+      startValue = 1;
+    }
```
nazarmatsevych, line 14: It's a good case to use ternary to set correct resultValue and startValue

nazarmatsevych (approved):

PR #1619, opened 2023-07-05
Solution

AnnaViolin23:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
```
Anastasiia-Svintsova, line 8: You could use a ternary operator to avoid `if` statement
```suggestion
    let accumulator = arguments.length < 2 ? this[0] : initialValue;
```

Anastasiia-Svintsova (approved):

Anastasiia-Svintsova (approved):

PR #1621, opened 2023-07-05
solution

Constantin-Haranskiy:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = initialValue;
+
+    if (this) {
```
etojeDenys: The condition 'if (this)' is redundant because 'this' always refers to the current object. It will always be true unless the function is called with null or undefined.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+      for (let i = 0; i < this.length; i++) {
+        if (i === 0) {
```
etojeDenys: The condition 'if (i === 0)' could lead to unexpected behavior. In the first iteration, 'i' is always equal to 0. The logic inside this condition will only be applied once. Moreover, if the initialValue is not provided, skipping the first element of the array is not a standard behavior of the 'reduce' method.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
+          arguments.length > 1
+            ? result = initialValue
+            : result = this[i++];
```
etojeDenys: Incrementing 'i' with 'i++' inside the assignment operation could result in skipping array elements during iteration. It is not recommended to manipulate the iterator inside the loop.

etojeDenys (changes requested):

TarasHoliuk (changes requested):
Hey! You haven't pushed any changes after the previous review. Push them to make the next review stage possible

PR #1629, opened 2023-07-08
add solution

denys-kononenko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator;
+    let i;
```
nazarmatsevych, line 9: We can add start values for accumulatorr and i variables and change them in if condition so we can avoid using else statement

nazarmatsevych (approved):

PR #1631, opened 2023-07-10
Solution

Lazarend:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,15 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let result = arguments.length > 1 ? initialValue : this[0];
+    const len = this.length;
```
Anastasiia-Svintsova, line 9: It's not necessary to move the length value to a variable, you use it only once

Anastasiia-Svintsova (approved):

PR #1633, opened 2023-07-11
Solution

Vitalii-Fediaiev:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,22 @@
...
+    if (arguments.length === 1) {
+      accumulator = this[0];
+
+      for (let i = 1; i < this.length; i++) {
+        accumulator = callback(accumulator, this[i], i, this);
+      }
+    } else {
+      for (let i = 0; i < this.length; i++) {
+        accumulator = callback(accumulator, this[i], i, this);
+      }
```
Anastasiia-Svintsova: You repeat the same logic in 2 cases. You can get rid of if/else statement by using a ternary operator.
```
const accumulator = ... ? ... : ... ;
const index = ... ? ... : ... ;

for (let i = index .....

**Review by Anastasiia-Svintsova — CHANGES_REQUESTED** [review]

File: src/arrayMethodReduce.js [hunk 2]
```diff
@@ -5,8 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
-  };
-}
+    let accumulator = arguments.length >= 2 ? initialValue : this[0];
```
```suggestion
    let accumulator = arguments.length === 2 ? initialValue : this[0];
```
Actually, there can be 2 arguments max

src/arrayMethodReduce.js
```diff
@@ -5,8 +5,16 @@
...
+    const startIndex = arguments.length >= 2 ? 0 : 1;
+
+    for (let i = startIndex; i < this.length; i++) {
+      accumulator = callback(accumulator, this[i], i, this);
+    }
 
+    return accumulator;
+  };
+  };
+
```
polosanya, line 18:
```suggestion
  }};
```

polosanya (approved):
Nice!

PR #1636, opened 2023-07-11
Added task solution

ruslan-ihnatenko:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this.length === 0) {
+      return initialValue;
+    }
+
+    let current;
+    let itr = 0;
```
polosanya, line 13: It would be great to use more meaningful name

polosanya (approved):
Good job!

PR #1637, opened 2023-07-13
implemented method reduce

Solovei-Andrii:


src/arrayMethodReduce.js
```diff
@@ -1,11 +1,20 @@
...
 
-/**
- * Implement method Reduce
- */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let acum = initialValue;
+
+    if (arguments.length === 1 && typeof this[0] === 'number') {
```
volodymyr-soltys97: The condition for checking the type of the first element in the array is not enough. It's not only numbers and strings that can be reduced, any type can be reduced. This can lead to unpredictable behavior.

volodymyr-soltys97 (changes requested):
Good job
But let's make your code perfect

OleziO (approved):

PR #1638, opened 2023-07-13
js_array-method-reduce

KarolinaWiktoria:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,15 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let initialValueCopy = arguments.length > 1 ? initialValue : this[0];
```
DorotaLeniecDev: For readability I would call this variable ```accumulator``` 😸

DorotaLeniecDev (approved):
Neat! One suggestion from my side but task is approved!

PR #1650, opened 2023-07-19
add task solution

andrew17752985:


src/arrayMethodReduce.js
```diff
@@ -5,8 +5,24 @@
...
+    let accumulator;
+    let startIndex;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      startIndex = 1;
+    } else {
+      accumulator = initialValue;
+      startIndex = 0;
+    }
```
VitaliyBondarenko1982:
```suggestion
  let accumulator = initialValue;
  let startIndex = 0;

    if (arguments.length < 2) {
      accumulator = this[0];
      startIndex = 1;
    }
```
better to avoid else

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):
Well done

nazarmatsevych (approved):

PR #1656, opened 2023-07-22
add task solution

vovaAlieksieiev:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const initualValuePassed = arguments.length > 1;
+
```
volodymyr-soltys97: The variable 'initualValuePassed' might be a typo. Rename it to 'initialValuePassed' to improve code readability.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

PR #1667, opened 2023-07-28
metodo riduci personalizzato

AndrzejPec:


src/arrayMethodReduce.js
```diff
@@ -3,9 +3,32 @@
...
-    // write code here
+    if (typeof callback !== 'function') {
+      throw new Error('Callback must be a function');
+    }
+
+    if (this.length === 0) {
+      return initialValue;
+    }
+
+    let startIndex = 0;
```
choeqq, line 17: It is good practice to initialize variables with empty values

src/arrayMethodReduce.js
```diff
@@ -3,9 +3,32 @@
...
+    let acc;
+
+    if (arguments.length === 2) {
+      acc = initialValue;
+    } else {
+      acc = this[0];
+      startIndex = 1;
+    }
+
+    for (let i = startIndex; i < this.length; i++) {
```
choeqq, line 27: It can also be done with recursion, but loop is also fine

choeqq (approved):
Well done ✅

PR #1668, opened 2023-07-29
Array_reduce solution

anpiet:


src/arrayMethodReduce.js
```diff
@@ -2,11 +2,56 @@
...
+const sourceStr = ['a', 'b', 'c', 'd'];
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+
+    const args = arguments.length;
+    const firstArg = this[0];
+
+    function valueCheck(value) {
```
choeqq, line 18: I think that chcecking this has length is good enough, but nice job with that

src/arrayMethodReduce.js
```diff
@@ -2,11 +2,56 @@
...
+    } else if (valueCheck(initialValue) === "noInitString") {
+      sum = "";
+    } else if (valueCheck(initialValue) === "noInitNumber") {
+      sum = 0;
+    } else if (valueCheck(initialValue) === "isValid") {
+      sum = initialValue || this[0];
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      sum = callback(sum, this[i], i, this)
```
choeqq, line 50: It can also be done using recursive function, but loop is also fine ✅

src/arrayMethodReduce.js
```diff
@@ -2,11 +2,56 @@
...
+
+const sourceNumbers = [0, 10, 20, 30];
+const sourceStr = ['a', 'b', 'c', 'd'];
+
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+
+    const args = arguments.length;
+    const firstArg = this[0];
```
choeqq, line 16: You can destructure both first and the rest of argumenty from this

choeqq (approved):
Nicely done, Please take a look at my comments 🚀

PR #1672, opened 2023-07-31
solution

gabriella135:


src/arrayMethodReduce.js
```diff
@@ -4,9 +4,21 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
```
loralevitska: The use of the rest parameters for the initial value is not needed. The second parameter to the 'reduce' function is not an array, but a single value. Replace '...initialValue' with 'initialValue'.

loralevitska (changes requested):
Almost done) Let's improve

Viktor-Kost (approved):

PR #1673, opened 2023-08-01
Develop

SichVictor:


OleziO (approved):

src/arrayMethodReduce.js
```diff
@@ -4,9 +4,15 @@
  * Implement method Reduce
```
OleziO: Remove all comments from the code.

OleziO (changes requested):

IvanFesenko (approved):
Well done

PR #1678, opened 2023-08-02
Reduce Method

Kacper-Leman:
soultion

### DIFF
--- a/src/arrayMethodReduce.js
+++ b/src/arrayMethodReduce.js
@@ -5,14 +5,12 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    let accumulator, startIndex;
+    let accumulator = initialValue;
+    let startIndex = 0;
 
-    if (initialValue === undefined && arguments.length < 2) {
-      accumulator = this[0];
+    if (arguments.length < 2) {
       startIndex = 1;
-    } else {
-      accumulator = initialValue;
-      startIndex = initialValue === undefined ? 0 : 0;
+      accumulator = this[0];
     }
 
     for (let i = startIndex; i < this.length; i++) {
--- a/src/arrayMethodReduce.js
+++ b/src/arrayMethodReduce.js
@@ -5,7 +5,21 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator, startIndex;
+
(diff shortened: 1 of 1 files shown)


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,21 @@
...
-    // write code here
+    let accumulator, startIndex;
+
+    if (initialValue === undefined && arguments.length < 2) {
+      accumulator = this[0];
+      startIndex = 1;
+    } else {
+      accumulator = initialValue;
+      startIndex = initialValue === undefined ? 0 : 0;
+    }
```
DorotaLeniecDev: This could be simplified a bit

```js
    let startIndex = 0;

    if (arguments.length < 2) {
      startIndex = 1;
      accumulator = this[0];
    }
```

DorotaLeniecDev (changes requested):
Good job, one improvement can be done :)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
+      startIndex = 1;
+      accumulator = this[0];
+    }
+
+    for (let i = startIndex; i < this.length; i++) {
+      accumulator = callback(accumulator, this[i], i, this);
```
AleksanderGelo, line 17: Ensure that the callback function in 'reduce2' has a return value. Without it, the behavior of 'reduce2' may become unpredictable and lead to errors.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
```
AleksanderGelo, line 7: Direct modifications to the Array prototype are generally a bad idea because they can lead to unpredictable behavior in the code. Consider creating a standalone function instead.

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
-    // write code here
+    let accumulator = initialValue;
+    let startIndex = 0;
+
+    if (arguments.length < 2) {
```
AleksanderGelo, line 11: Use more descriptive condition check rather than checking arguments length. You can directly check if 'initialValue' is undefined.

AleksanderGelo (approved):

PR #1684, opened 2023-08-08
add task solution

Wesses:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let i = 0;
```
davinnchii: I think you should change name to startIndex or something like in MDN doc
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce)

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
+
+    if (arguments.length === 1) {
+      accumulator = this[0];
+      i++;
+    }
+
+    for (; i < this.length; i++) {
```
davinnchii: In my opinion it will be better to write `i` before ; in this line

davinnchii (changes requested):
LGTM, but need some changes

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
-    // write code here
+    let accumulator = initialValue;
+    let index = 0;
+
+    if (arguments.length === 1) {
+      accumulator = this[0];
+      index++;
+    }
+
+    for (; index < this.length; index++) {
```
33Orange, line 16:
```suggestion
    for (index; index < this.length; index++) {
```

33Orange (approved):
Well done

PR #1685, opened 2023-08-08
Add task solution

davinnchii:


src/arrayMethodReduce.js
```diff
@@ -4,9 +4,24 @@
...
+    let accumulator = initialValue.length > 0
+      ? initialValue[0]
+      : this[0];
+    const startIndex = initialValue.length ? 0 : 1;
+
+    if (startIndex === 1 && this.length > 0) {
+      accumulator = this[0];
+    } else if (startIndex === 0 && initialValue === undefined) {
+      accumulator = 'undefined';
+    }
```
Wesses: can be done without else if. Just rotate accumulator = 'undefined' before if.

src/arrayMethodReduce.js
```diff
@@ -4,9 +4,24 @@
...
+
+    for (let i = startIndex; i < this.length; i++) {
```
Wesses: U have created the variable for index, `startIndex`. but it's up to you

Wesses (changes requested):
Just a few changes needed)

src/arrayMethodReduce.js
```diff
@@ -4,9 +4,18 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
+    let accumulator = initialValue.length > 0
+      ? initialValue[0]
+      : this[0];
+    let startIndex = initialValue.length ? 0 : 1;
```
33Orange, line 11: The ternary operator usage for setting startIndex can lead to incorrect results if initialValue is undefined. You need to check if initialValue is not undefined before checking its length.

src/arrayMethodReduce.js
```diff
@@ -4,9 +4,18 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
```
33Orange, line 7: The usage of the ... operator can lead to incorrect results if more than one initial value is passed. It's better to use two separate arguments instead: function(callback, initialValue)

src/arrayMethodReduce.js
```diff
@@ -4,9 +4,18 @@
...
+    let accumulator = initialValue.length > 0
```
33Orange, line 8: The ternary operator usage for checking the length of initialValue can lead to incorrect results if initialValue is undefined. You need to check if initialValue is not undefined before checking its length.

33Orange (changes requested):

src/arrayMethodReduce.js
davinnchii: @33Orange we discussed it with @alexander-ignatow  on Q&A session, that this is the most proper solution, because for tests we need to be able to reduce the array if initial value was set manually to undefined. Only 2 solutions we got are `arguments` or making initialValue parameter to be an array. As I remember coach said that arguments is distant solution and it is better not to use it, but I have to complete that test with undefined, so. If you think that proper decision is using arguments - I will refactor it.

BudnikOleksii (approved):
Good job!

src/arrayMethodReduce.js
BudnikOleksii: @davinnchii  Don't forget to answer comments if you don't agree.

src/arrayMethodReduce.js
BudnikOleksii: Yes, it's better to use the rest operator instead of arguments, but the native method accepts only 2 arguments and uses them.

PR #1688, opened 2023-08-08
task solution

viktoria-roik:


maksym-mishchanchuk (approved):
likeit

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
...
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accumulator = initialValue;
+    let startedIndex = 0;
+
+    if (arguments.length < 2) {
+      accumulator = this[0];
+      startedIndex = 1;
+    }
```
tiserett: let's use ternary here

tiserett (changes requested):

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const argumentsLengthCompare = arguments.length < 2;
```
Moroz-Dmytro, line 8: But it returns `boolean` so I'd call it `hasInitialValue`, in this case when we add `has` we show that this is `boolean` variable

Moroz-Dmytro (approved):
Good work 🚀 
Small comment below

PR #1690, opened 2023-08-08
initial commit

maksym-mishchanchuk:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,20 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, props) {
```
alexander-ignatow: `props` is not perfect name for the variable
The meaning of it is to pass the initial value to accumulator - you need to pick something descriptive

alexander-ignatow (changes requested):
Almost done!

Please fix minor issue

polinavafik (approved):
Well done!✨

I would rename `count` to something else like `startIndex`, but it's up to you!

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    let accum = initialValue;
+    let count = 0;
```
BudnikOleksii, line 9: Agree with the previous reviewer, "count" isn't informative

BudnikOleksii (approved):
Good job!
Pay attention to reviewers' comments.

PR #1695, opened 2023-08-10
Solved task

AndriiKuz:


desevoker (approved):
Good job! 🚀

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,15 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
```
33Orange: You are spreading the initialValue argument. The second argument to reduce should not be an array but a single value. Remember that reduce takes a callback and an optional initialValue as arguments.

To check the length of arguments you can use a [`arguments`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) inside the function

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #1696, opened 2023-08-10
add task solution

P-Nazar:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    if (this.length === 0 && initialValue === undefined) {
+      return undefined;
+    }
```
Viktor-Kost, line 10: this condition is redundant
```suggestion
```

src/arrayMethodReduce.js
```diff
@@ -5,7 +5,23 @@
...
+
+    let startIndex = 0;
+    let accumulator = initialValue;
+
+    if (arguments.length < 2) {
+      startIndex = 1;
+      accumulator = this[0];
+    }
```
Viktor-Kost, line 18: it's better to use ternary to set correct values

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

PR #1699, opened 2023-08-11
add task solution

vserhiichuk:


src/arrayMethodReduce.js
```diff
@@ -4,8 +4,15 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
+    let accumulator = initialValue.length === 0 ? this[0] : initialValue[0];
```
kaavka: We could use arguments.length here, so we wouldn't need to use the rest operator.

kaavka (approved):
Nice and clean ✨

src/arrayMethodReduce.js
```diff
@@ -4,8 +4,15 @@
  * Implement method Reduce
  */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
```
33Orange: Don't use '...initialValue' as it can take multiple arguments, but according to the native 'reduce' method, it should take only one argument after the callback.

src/arrayMethodReduce.js
33Orange:
```suggestion
    let accumulator = arguments.length === 0 ? this[0] : initialValue[0];
```

33Orange (changes requested):
Almost done

tiserett (approved):

PR #1700, opened 2023-08-11
add solution

KatOlista:


cty2802 (approved):
well done!

src/arrayMethodReduce.js
```diff
@@ -1,11 +1,16 @@
...
-/**
- * Implement method Reduce
- */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
+    let [accum] = initialValue.length ? initialValue : [this[0]];
+
+    let i = accum === this[0] ? 1 : 0;
```
tiserett: create startIndex variable, no need to put 'i' outside of loop

src/arrayMethodReduce.js
```diff
@@ -1,11 +1,16 @@
 'use strict';
 
-/**
- * Implement method Reduce
- */
 function applyCustomReduce() {
-  [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+  [].__proto__.reduce2 = function(callback, ...initialValue) {
```
tiserett, line 4: not sure if you should be using rest here

tiserett (changes requested):

src/arrayMethodReduce.js
KatOlista: i have no info that i shouldn't

Moroz-Dmytro (approved):

PR #1713, opened 2023-08-15
add task solution

k1NAs:


src/arrayMethodReduce.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomReduce() {
   [].__proto__.reduce2 = function(callback, initialValue) {
-    // write code here
+    const ZERO = 0;
+    const ONE = 1;
```
Moroz-Dmytro, line 9: I guess, these constants are useless in this case. You don't really need them

Moroz-Dmytro (approved):
LGTM 👍 
Only small comment

33Orange (approved):

